PR #45, opened 2013-05-06
Update logging.py

spmorton:
line 416 - argv[0].rsplit('/', 1)[1] , the split would cause job crashes that did not provide a splittable input. Not sure why you would want to split argument 0 here anyway. Offending script was mythvidexport.py. I know it is not a perfect solution, but making a check for the possibility of an out of index error in a cursor call seemed a little more extensive than I cared to contribute just for a log entry


rcrdnalor:
This was fixed by commit 246114b on  Aug 24, 2013 .

PR #62, opened 2014-01-06
WebFrontend: Allow the main pages to be bookmarked

martinjt:
This sets up the ability to bookmark some pages in the WebFrontend, along with the infrastructure to do more in the future.


stuarta:
I'm going to close this as we are replacing this entire code with a new implementation

PR #69, opened 2014-03-05
Updated getKeys() and getHosts() xml bindings to latest format

fracmak:
Was playing around with python bindings and noticed that the getHosts() and getKeys() functions no longer worked due to 'Values' not being present in the resulting feed. This removes them to restore functionality


stuartm:
Committed to master

PR #70, opened 2014-03-09
Services API - Get cut list and comm break for recorded

janbar:
This is a patch adding the ability to get cut list and comm break for recordings.
Also new servives are able to return offset by keyframe, position or duration
depending of request parameter "OffsetType" { nil , "Position", "Duration" }


janbar:
Hi @stuartm , i rebased since many users want to have it. br

stuartm:
Committed to master

PR #73, opened 2014-05-01
Fix for youtube crash

jlazkano:
Here is more information: http://www.gossamer-threads.com/lists/mythtv/users/513770


stuarta:
I've reviewed this and this fix was already included in commit https://github.com/MythTV/mythtv/commit/07eff1b411d190a22671df10dce8e903ef73cd95

Thank you for your contribution.

PR #75, opened 2014-06-14
Use a separate thread for recursive directory scans

chadparry:
Slideshows can take a long time to start if there are lots of files and directories in the album tree. The solution is to scan the album on a different thread while the slideshow is already running. This makes the `imageview` more complicated, because of the inter-thread coordination. Fortunately the changes can mostly be contained in the `imageview`. The benefits to the user are noticeable.


dekarl:
This was committed in January for 0.28.

PR #88, opened 2014-11-22
Add UPC Cablecom subtitle support

ReArtu:
I hope this is the right way to do it. It works for me but I don't know if it mess with other provider. (This change liked to dvbdescriptor.h). Hope it's clean enough for you...


ReArtu:
Really sorry for this late answer... so much work to do.
It works like a charm. Thank you very much !

PR #89, opened 2014-12-17
Random slideshow with a seasonal theme

chadparry:
This creates a new mode of slideshow, called "Seasonal," to supplement the existing sequential and random slideshows. It shows a random slideshow but with a larger weight given to photos from the same time of year. The feature is only enabled if EXIF support is present. I also made some fixes to the sequence classes, so that they tracked history better and could run forever without leaking memory.

I wrote this on top of a related feature, (https://github.com/MythTV/mythtv/pull/75), because otherwise pulling both would cause merge conflicts.


dekarl:
This was committed in December/January to make it into 0.28

PR #90, opened 2015-01-02
Add support for DVB-T2 for tuners that do not auto-switch

dcjm:
This patch provides support for DVB-T2 in the same way as DVB-S2 is supported.  The drivers for the first DVB-T2 tuners (e.g. PCTV-290)  automatically switched between DVB-T and DVB-T2 so that there was no need for the application to treat it specially.  More recent drivers, such as that for the Si2168 in the DVBSky T9580 and PCTV 292e require explicit switching.

There are separate commits for the basic support and for channel scanning.  The channel scanner includes support for editing transports and for a full scan.  The full scan tries to lock each frequency with DVB-T and only tries DVB-T2 if that fails.  This avoids creating duplicate entries when scanning with a device that ignores the DVB-T/T2 setting and locks to either.  There is currently no support for PLPs (physical layer pipes).

### DIFF
--- a/mythtv/libs/libmythtv/cardutil.cpp
+++ b/mythtv/libs/libmythtv/cardutil.cpp
@@ -431,9 +431,13 @@ QString CardUtil::ProbeDVBType(const QString &device)
 
     DTVTunerType type(info.type);
 #if HAVE_FE_CAN_2G_MODULATION
-    if (type == DTVTunerType::kTunerTypeDVBS1 &&
-        (info.caps & FE_CAN_2G_MODULATION))
-        type = DTVTunerType::kTunerTypeDVBS2;
+    if (info.caps & FE_CAN_2G_MODULATION)
+    {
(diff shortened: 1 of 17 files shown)


dekarl:
This PR is being tracked as https://code.mythtv.org/trac/ticket/12342

stuarta:
This has now been merged to master

notroj:
Has anybody tried backporting this set of changes to fixes/0.27, out of interest?

dcjm:
Yes, there are three commit in [https://github.com/dcjm/mythtv/tree/fixes/0.27](https://github.com/dcjm/mythtv/tree/fixes/0.27).  Admittedly, this was over a year ago so I don't know if they'll apply cleanly to the current state.

stuarta:
I had originally planned to backport this to 0.27, but there were some issues that took a while to be resolved. By the time they were resolved, we were working flat out to get 0.28 released.

Not sure yet if it will eventually get merged.

PR #92, opened 2015-03-03
fix SSDP thread using 100% CPU on zero byte datagrams

warrenfalk:
Fix for ticket #12395 (submitted by me). Fixes a denial of service caused by a zero byte datagram to the SSDP port (which is actually happening right now by at least one of my other devices, presumably by accident).


stuartm:
Merged to 0.27/fixes and master

PR #93, opened 2015-03-17
Pull request for 11491 (Add querying of recorded markup data to Services API)

dhaber:
Update https://code.mythtv.org/trac/ticket/11491 to be in sync with current Master.

Also added Bookmark property


dekarl:
Closing as this is going to be solved differently. Quoting from https://code.mythtv.org/trac/ticket/11491#comment:8

> Although there were overlaps, some of the functionality provided by this patch is not provided by the patches in #12090.
>
> This patch allows pulling the bookmark information and information about resolution, bitrate and aspect ratio.
>
> Which isn't to say that this patch will be merged, it won't, but I'm leaving it open as a reminder to add the equivalent functionality to the services API. Namely the ability to pull recording file metadata - resolution, bitrate, codec etc from the new recordedfile table.

PR #95, opened 2015-05-03
Fixes for visualization under 24-bit and 32-bit audio

pHr34kY:
I noticed that the visualizer was rendering garbage whenever I played a vorbis file or a 24-bit flac. It seems that the visualizers require a 16-bit stereo feed, and currently only convert between 8 and 16 bit (mono and stereo). The 24-bit FLAC decoder outputs 32bit signed integers, and the Ogg Vorbis decoder outputs 32 bit float (even for stuff encoded from a 16-bit source). The only way to pick the difference between the two (and choose the correct cast) is that FLAC reports 24-bit, even though both are outputting 32.

I have also stopped it from rendering uninitialised memory for an unsupported sample format, the same way it does when there is an unsupported channel count.

Note: Raised on trac as ticket #12445

### DIFF
--- a/mythplugins/mythmusic/mythmusic/inlines.h
+++ b/mythplugins/mythmusic/mythmusic/inlines.h
@@ -81,6 +81,30 @@ static inline void stereo16_from_stereopcm16(register short *l,
 }
 
 
+static inline void stereo16_from_stereopcm32(register short *l,
+                         register short *r,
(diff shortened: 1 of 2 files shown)


dekarl:
Closing PR as this has already been committed as b6331d71e08a85999f24f4810c5a8344b973f934

PR #96, opened 2015-07-31
Fix includes for qt-5.5 compatibility

martyg77:
Fix broken trunk build when compiling against qt-5.5
Ref. Ticket 12480


stuarta:
commited. Thanks very much

PR #98, opened 2015-10-07
Fix mytharchive

benklop:
in current mytharchive, using python 2.7.9, mythburn.py fails due to not properly formatting the hostname into a query retrieving settings from the db. this should fix that issue.


dekarl:
thank you for the contribution, fixed in 9ce6aaa2f08d57a86068e65c6078492b03880b23

PR #103, opened 2015-12-07
Add heuristic in TMDB script to detect movie release years

chrisaljoudi:
This enables movies named per [MythVideo's naming conventions](https://www.mythtv.org/wiki/MythVideo_File_Parsing) to succeed with TMDB metadata searches.

### DIFF
--- a/mythtv/programs/scripts/metadata/Movie/tmdb3.py
+++ b/mythtv/programs/scripts/metadata/Movie/tmdb3.py
@@ -29,6 +29,7 @@ __version__ = "0.3.7"
 
 from optparse import OptionParser
 import sys
+import re
 
 def buildSingle(inetref, opts):
     from MythTV.tmdb3.tmdb_exceptions import TMDBRequestInvalid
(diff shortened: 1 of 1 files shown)


dekarl:
I take it.
Personally I'm in favour of unifiying behaviour between recordings and videos.
Will this patch find movies where the year is off by a bit? e.g. by looking at all release events of a movie. Will this work for e.g. "Blues Brothers 2000"?
Generally we ask to have all contributions registered on trac so, they do not get overlooked when coming down the various channels.

chrisaljoudi:
@dekarl Sorry for the late response — been a bit busy.

> Will this patch find movies where the year is off by a bit?

Probably not (it's dependent on what TMDB returns). What this patch does in a nutshell:

* Say we're searching for movie with file name `str`.
* If `str` ends with a year, extract it to `year` and remove it from `str`
* Use `str` to search TMDB just as you normally would, BUT:
* Also passing `year` to the TMDB API as a "year" criterion (if we have a year)

There's already [some dead code](https://github.com/MythTV/mythtv/blob/e7e9a937c5a66b412a7f54a404894cc5bcc5b5f7/mythtv/bindings/python/tmdb3/tmdb3/tmdb_api.py#L132) (dead AFAICT), which comprises a function `searchMovieWithYear` — that function looks for a year of the form `(yyyy)` at the end of the search string and, if it finds one, includes it in the TMDB API search request.

However, it does not look like `searchMovieWithYear` is ever called. Even if it were called, the year syntax it strictly expects [contradicts the MythVideo file naming patterns as described in the Wiki](https://www.mythtv.org/wiki/MythVideo_File_Parsing).

Note that this patch *optionally* accepts parenthesis around the year.

> Will this work for e.g. "Blues Brothers 2000"?

Doesn't look like it. The detected release year we ask TMDB to filter by is 2000. That said, it's worth noting this is a far less frequently-occurring case than otherwise.

---

The only reasonable approach I can think of to solve that problem (and many others), and relating to what you said...

> unifiying behaviour between recordings and videos

Right, a major overhaul of the metadata lookup infrastructure is the way to do this if we want to fix edge cases like "Blues Brothers 2000" (and many others) and make the lookup smarter and more resilient overall.

Obviously, that's more work, and while I'd be happy to spend time on it (let me know your thoughts), it seems like what **this** patch addresses is an easy, low-hanging fruit to significantly improve success rates with movie metadata lookup.

chrisaljoudi:
@dekarl Let me know your thoughts.

chrisaljoudi:
@dekarl Ack?

dekarl:
Happy new Year! I've been thinking about this on and off.
I do agree that the documented file names should work, even though I think a hard match on exact year is not of much use against themoviedb. But I think the low hanging fruit would be to do two searches, one for "Blues Brothers", "year 2000" and another one for "Blues Brothers 2000". Maybe only do the second if the first doesn't find anything, maybe just do both and combine the results. Can you add one of the two variants?
The rewrite for better automated matching, e.g. taking movie length into account to know if its "Pixels" the short or "Pixels" the Hollywood movie, needs some more thought. Is there useful metadata in the video files that can be used for matching? e.g. cast, director, date of release, duration.

rcrdnalor:
The handling of file names ending with a year was updated by #975 in a slightly different way:
The following patterns are now allowed and are honored by the movie/tv grabbers:
- A Movie Title (YYYY).mpg
- Title/Season 1/Title (YYYY) s01e01 Subtitle.mpg
See [MythVideo File Parsing](https://www.mythtv.org/wiki/MythVideo_File_Parsingl) on the wiki
PR closed

PR #104, opened 2015-12-18
Proposed additions/updates to MythTV

neoh4x0r:
1) Updated twit.tv.pl internetcontent script.
 - fixes scraping of videos and episodes (only grabs active shows)
 - uses the wiki's rss feed instead of the main site (faster and is less of a burden on their main site)
2) Updated theme url (to http://themes.mythtv.org/themes/repository/trunk/themes.zip)
- for 0.28-pre the theme url did not exist: http://themes.mythtv.org/themes/repository/0.28-pre/themes.zip
3) Commented out some unnecessary osd messages about tuning/lock timeout in tv_rec::TuningSingalCheck
- I use composite input which does not need to tune to a channel
- the osd message must be exited out of before you can do anything with the interface and is generally not needed
- note: it still logs these to the console/logfile/database if active

### DIFF
--- a/mythtv/libs/libmythtv/tv_rec.cpp
+++ b/mythtv/libs/libmythtv/tv_rec.cpp
@@ -3899,18 +3899,18 @@ MPEGStreamData *TVRec::TuningSignalCheck(void)
             newRecStatus = RecStatus::Failing;
             curRecording->SaveVideoProperties(VID_DAMAGED, VID_DAMAGED);
 
-            QString desc = tr("Good signal seen after %1 ms")
-                           .arg(genOpt.channel_timeout +
-                        startRecordingDeadline.msecsTo(MythDate::current()));
+//            QString desc = tr("Good signal seen after %1 ms")
+//                           .arg(genOpt.channel_timeout +
+//                        startRecordingDeadline.msecsTo(MythDate::current()));
             QString title = curRecording->GetTitle();
             if (!curRecording->GetSubtitle().isEmpty())
                 title += " - " + curRecording->GetSubtitle();
 
-            MythNotification mn(MythNotification::Check, desc,
-                                "Recording", title,
-                                tr("See 'Tuning timeout' in mythtv-setup "
-                                   "for this input."));
-            gCoreContext->SendEvent(MythEvent(mn));
+//            MythNotification mn(MythNotification::Check, desc,
+//                                "Recording", title,
+//                                tr("See 'Tuning timeout' in mythtv-setup "
+//                                   "for this input."));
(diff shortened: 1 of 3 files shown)


dekarl:
Are there already tickets for the three issues?
* twit.tv.pl update
* an OSD message appearing for a tuning timeout on recorders that do not tune (did I understand that correctly?)
* The issue with the theme downloader not using trunk on master
After a quick look the two latter patches appear to be hacks that break functionality down the road / in other use cases.

neoh4x0r:
I did do a search for the issues (as well as following the guidelines for checking if a ticket already exists) and I could not find any (open/closed).

Yes  the tuning timeout message is displayed anytime you attempt to start live tv (or change channels) -- even if the input source is not tuned (eg composite/s-video and the like).

The theme url does not exist for v0.28-pre (http://themes.mythtv.org/themes/repository/0.28-pre/themes.zip) whereas the url v0.27.5 does exist.

Also, the twit.tv.pl script has been broken since at least 0.27.5 (maybe even earlier) since the twit.tv website was modified (but they do have standard rss 2.0 feeds on the wiki).

dekarl:
Ok. Can you please open a ticket for each issue so the relevant developers can take a look at it.
https://code.mythtv.org/trac/wiki/TicketHowTo#OpeningaTicket

neoh4x0r:
@dekarl  -- I will open tickets for each of those issues.
It might be a little while before I can do that -- I am reverting my system back to Debian/Jessie after having tried out Stretch and Sid (too many broken packages to count).

neoh4x0r:
@dekarl : I have justed opened the tickets on the trac server

#12571 (Theme url does not exist for lastest git version 0.28-pre)
https://code.mythtv.org/trac/ticket/12571

#12572 (Non-tunable inputs always generate tuning timeout message spam when watching live tv)
https://code.mythtv.org/trac/ticket/12572

#12573 (Update to Mythnetvision twit.tv script by Auric)
https://code.mythtv.org/trac/ticket/12573

neoh4x0r:
in regards to bug #12571 -- (https://code.mythtv.org/trac/ticket/12571)
the root cause of the url mismatch was because the .git folder was missing (it got deleted).
That meant that version.sh had defined MYTH_SOURCE_PATH as an empty string (instead of master).
in themechooser.cpp it expects the branch to be master or the url is built using MYTHTV_BINARY_VERSION (which is why the url did not exist 0.28 has not been released yet). 

Also, bug #12572 (https://code.mythtv.org/trac/ticket/12572)
is still open (and was accepted by jpoet) -- I'm not sure of what the status is on it.

PR #105, opened 2015-12-18
Updated British English translations for Gallery

knowledgejunkie:


dekarl:
This is already being tracked over on Trac, see https://code.mythtv.org/trac/ticket/12569

PR #113, opened 2016-03-04
Add a new default video display profile called OpenMAX Normal for raspberry pi

bennettpeter:
Raspberry pi needs to use OpenMAX decoder and renderer. There should be a default display profile to assist in the frontend video playback setup. This pull request adds that to the frontend setup.

Ticket for this pull request:
https://code.mythtv.org/trac/ticket/12675


dekarl:
Fixed in 5450aa53d726708333da1e896d73876f3926a916 and 5d06abc5fd601ed955bbf9c0a5f70b039b52e9f7.

PR #114, opened 2016-03-07
Pullreq raspi OSD fixes

bennettpeter:
When playing recordings or videos in HD on raspberry PI with xcb, if you do something that displays an OSD, audio becomes choppy and video becomes jerky until the OSD is taken down. The only OSD option for raspberry pi with xcb is softblend. These fixes optimize the softblend process in several ways. 
1. Prevent blending OSD into frames that will be skipped.
2. Multi threading designed by LVR to speed the OSD blending.
3. Eliminate merging of areas of the OSD that are either opaque or invisible. For invisible areas: skip, for opaque areas: overwrite, instead of existing code that calculates a merge every time.
Performance with OSD is now almost perfect.
I recommend turning off "OSD fade", as that still causes a small glitch.
Ticket here: https://code.mythtv.org/trac/ticket/12677

### DIFF
--- a/mythtv/libs/libmythtv/DVD/mythdvdplayer.cpp
+++ b/mythtv/libs/libmythtv/DVD/mythdvdplayer.cpp
@@ -234,13 +234,6 @@ void MythDVDPlayer::DisplayLastFrame(void)
     SetScanType(kScan_Progressive);
     DisplayDVDButton();
 
-    osdLock.lock();
-    videofiltersLock.lock();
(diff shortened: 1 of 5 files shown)


dekarl:
Fixed in 7f8614ec, 109bf9d38983cb764d6fa9ea1cf34b16898efdfb and 5d06abc5fd601ed955bbf9c0a5f70b039b52e9f7

PR #115, opened 2016-03-20
Fix bleed through of background when playing video with OpenMAX

bennettpeter:
This fixes the bug in ticket: https://code.mythtv.org/trac/ticket/12691
Fixes:
1. With OpenMAX playback profile if you play a video that is letterbox or pillarbox, the blank areas next to the video shows artifacts from the playback window, depending on the theme.
2. When changing resolution the screen displays "Please Wait", or the whole playback window, depending on the theme

These are fixed by creating a video background window that is black if OpenMAX is being used.


dekarl:
Fixed in aa676f77d

PR #117, opened 2016-04-11
Added dvr services call for setting and getting bookmark position

mitchcapper:
This PR adds two functions GetSavedBookmark and SetSavedBookmark they were done largely in the style of #12090. I thought about adding the ability to pull the full markup (in fact I am not sure the other GetRecordedCutList shouldn't do this and let the user sort the types out) but this still left how best to set a bookmark without adding a more complex markup setting function.
I think this simplified version of just Get/Set bookmark adds the remaining  missing hunk to the services api for markup, no it doesn't add the metadata of #11491 but that probably should come from elsewhere.
Both functions take very similar args to #12090.
I did refactor the QueryKeyFramePosition and QueryKeyFrameDuration main function code into one function, as I was adding two more (QueryPositionKeyFrame,QueryDurationKeyFrame) and they were nearly identical I figured I would slim down the code.  If you prefer 4 functions with duplicate code however I can update the code for that.

### DIFF
--- a/mythtv/libs/libmythservicecontracts/services/dvrServices.h
+++ b/mythtv/libs/libmythservicecontracts/services/dvrServices.h
@@ -45,11 +45,12 @@
 class SERVICE_PUBLIC DvrServices : public Service  //, public QScriptable ???
 {
     Q_OBJECT
-    Q_CLASSINFO( "version"    , "6.1" );
+    Q_CLASSINFO( "version"    , "6.2" );
     Q_CLASSINFO( "RemoveRecorded_Method",                       "POST" )
     Q_CLASSINFO( "DeleteRecording_Method",                      "POST" )
     Q_CLASSINFO( "UnDeleteRecording",                           "POST" )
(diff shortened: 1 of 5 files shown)


mitchcapper:
Ticket created https://code.mythtv.org/trac/ticket/12720
#12720

mythtv/libs/libmythservicecontracts/services/dvrServices.h
```diff
@@ -45,11 +45,12 @@
 class SERVICE_PUBLIC DvrServices : public Service  //, public QScriptable ???
 {
     Q_OBJECT
-    Q_CLASSINFO( "version"    , "6.1" );
+    Q_CLASSINFO( "version"    , "7.0" );
```
billmeek: Mitch,

I got the version wrong, it should be 6.2 (stuartm answered my question on IRC)

mitchcapper:
fixed bill.

PR #120, opened 2016-05-04
PR for Services API Adding GetKeyBindList and PutKeyBind to Myth Services

mitchcapper:
This adds two service API calls GetKeyBindList and PutKeyBind.  The code here is largely modeled after the existing GetSettingList/PutSetting and storageGroupDir / storageGroupDirList code.

This is to make working with the new Frontend/SendKey function easier, so you can determine what key is mapped to what, and adjust key bindings. It doesn't notify/reload the frontend (but neither does MythWeb).

There is at least two major items for review/comment:
-PutSetting only returns true if it has changed the key binding.  It returns false if the key binding is invalid, or if it is already set to that key.  I could just have it always return true without checking if any rows are updated, or if we want to be very correct I could first do a select on the key to make sure it exists, and return true as long as it exists.

-I merged jumppoints and key bindings together (from a function point of view).  Jumppoints now show up as Context=JumpPoint.  Its a hack but it avoids having to make two more functions and more round trips for a client trying to get all the current binds.  If it is felt two more functions is better I can split it.

-Also I debated added an IsAlive type option to only return keybindings for hosts that are active (when hosts are not specified) I don't feel this is really needed however.


mitchcapper:
https://code.mythtv.org/trac/ticket/12753

mitchcapper:
https://code.mythtv.org/trac/ticket/12803

PR #121, opened 2016-06-06
Added ChildID to VideoService VideoMetadataInfo for Video Services API

mitchcapper:
Simple patch adding ChildID to the VideoMetadataInfo so that people who have multi-part videos have a way of re-assembling via the service API.


mitchcapper:
https://code.mythtv.org/trac/ticket/12803

PR #122, opened 2016-06-21
Fix segmentation fault in QMetaObject::indexOfClassInfo() #12782

PlanetEater:
The NULL check for pMetaProp in xmlSerializer.cpp should protect the GetContentName() call too.
Related (already existing) ticket: https://code.mythtv.org/trac/ticket/12782

### DIFF
--- a/mythtv/libs/libmythupnp/serializers/xmlSerializer.cpp
+++ b/mythtv/libs/libmythupnp/serializers/xmlSerializer.cpp
@@ -120,14 +120,15 @@ void XmlSerializer::AddProperty( const QString       &sName,
 {
     m_pXmlWriter->writeStartElement( sName );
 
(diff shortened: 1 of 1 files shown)


stuarta:
How does this fix the segfault? The code you have changed already does a check that pMetaProp != NULL, so I can't see how this actually fixes the issue?

PlanetEater:
The change is in what happens when pMetaProp == NULL.  Easy to overlook, apparently. :-)  I explained this a bit more on the ticket discussion.

PlanetEater:
Redid it in a way that's more self-explaining (I hope).

stuarta:
Committed to master in c3a7929853026d98fb85bf76a79061d6b02f9b47

PR #123, opened 2016-08-07
Update list of mark types in python bindings

tomhughes:
Resynchronises the list of mark types in the python bindings with the list in the C header file by adding missing types.


tomhughes:
Opened https://code.mythtv.org/trac/ticket/12849 to track this.

PR #125, opened 2016-11-26
add compatibility for libcec4

BlackIkeEagle:
https://code.mythtv.org/trac/ticket/12939#ticket

added updates to the libcec4 interface, so mythtv can be compiled
against it. Some parts might need some cleaning.

Signed-off-by: BlackEagle <ike.devolder@gmail.com>


stuarta:
Merged, thanks

PR #128, opened 2017-02-10
Update locale.py

Cypresslin:
Country name correction for future localization task.


dekarl:
Link back to trac https://code.mythtv.org/trac/ticket/12997

linuxdude42:
The official ISO website says that the current value is correct.

PR #133, opened 2017-04-02
Prevent duplicate case when ESTRPIPE == EPIPE

RealDeuce:
Fixes https://code.mythtv.org/trac/ticket/13023#ticket

### DIFF
--- a/mythtv/libs/libmythtv/recorders/audioinputalsa.cpp
+++ b/mythtv/libs/libmythtv/recorders/audioinputalsa.cpp
@@ -271,9 +271,11 @@ int AudioInputALSA::PcmRead(void* buf, uint nbytes)
                     break;
                 case -EINTR:
                 case -EPIPE:
+#if ESTRPIPE != EPIPE
                 case -ESTRPIPE:
                     Recovery(nread);
                     break;
+#endif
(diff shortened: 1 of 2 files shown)


mythtv/libs/libmythtv/recorders/audioinputalsa.cpp
```diff
@@ -271,9 +271,11 @@ int AudioInputALSA::PcmRead(void* buf, uint nbytes)
                     break;
                 case -EINTR:
                 case -EPIPE:
+#if ESTRPIPE != EPIPE
                 case -ESTRPIPE:
                     Recovery(nread);
                     break;
+#endif
```
stuarta: This will cause EPIPE not to be handled correctly. The `#endif` should be moved up so that it only wraps
`case -ESTRPIPE:`

linuxdude42:
If you fix the conflict, I'll get this committed.

stephen-hurd:
Sorry it took so long to fix the conflict, I moved across the country, and just got my MythTV box set up again.

linuxdude42:
One more request. Could you rebase this off current master so that its not a seven month old branch. Thanks.

PR #135, opened 2017-04-15
Devel/dh/nodebt

linuxdude42:
Changes from Qt3 to Qt4 removed the need to explicitly call the detach or copy functions on qt classes. These will be called internally as necessary. Fixes https://code.mythtv.org/trac/ticket/13026


linuxdude42:
This was committed via a different process.

PR #138, opened 2017-06-24
Consolidated DVB SI rework

rogerjames99:
Change name of SatelliteDeliverySystemDescriptor::FrequencyHz()  to FrequencyKHz and fix the places where the return value is incorrectly treated. This fixes the a problem with fully tuned scans of satellite delivery systems which would fail to tune any new muliplexes found due to incorrectly dividing the frequency got from the the descriptor by 1000.

Also dvbsi changes as a consolidated patch.

### DIFF
--- a/mythtv/.gitignore
+++ b/mythtv/.gitignore
@@ -6,3 +6,9 @@ Makefile.*.debug
 config.*
 version.cpp
 TMPMAK
+.project
+.cproject
+.settings
+logs
+errs
+src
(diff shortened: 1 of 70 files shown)


rogerjames99:
Added the dvbsi stuff as a consolidated patch.

rogerjames99:
I admit the dvbsi stuff got tagged onto this by accident. I thought I was creating a new pull request, but github obviously only allows one.

rogerjames99:
Closed tidyness. Will reopen against next release.

PR #142, opened 2017-08-14
 Handle Non Integer Refresh Rates with XRandr (Fixing JudderFree) 

halovanic:
The current code for [​JudderFree](https://www.mythtv.org/wiki/User_Manual:JudderFree) doesn't work anymore for ​[recent NVIDIA drivers](https://lists.gt.net/mythtv/users/597082#597082). It also was an NVIDIA-only solution due to relying on getting and setting refresh rates using XRandr, where they can only be integer values, or in NVIDIA's case, fake integer values.

This pull request uses an alternate method of calculating the actual refresh rate as a more accurate double value and switching to a supported modeline without trying to match the real and fake rates through the (broken) NVIDIA extension method. This should also allow Judder Free playback for other video cards so long as they support suitable implementation of XRandr.

Also attached in trac ticket [13100](https://code.mythtv.org/trac/ticket/13100)

### DIFF
--- a/mythtv/libs/libmythui/DisplayResX.cpp
+++ b/mythtv/libs/libmythui/DisplayResX.cpp
@@ -12,7 +12,11 @@
 
 #include <X11/extensions/Xrandr.h> // this has to be after util-x11.h (Qt bug)
 
-static XRRScreenConfiguration *GetScreenConfig(MythXDisplay*& display);
+static XRRScreenResources *GetScreenResources(MythXDisplay*& display);
+static XRROutputInfo *GetOutputInfo(MythXDisplay*& display, XRRScreenResources*& rsc);
+RRCrtc crtc;
+/* Width x Height x Resolution = X mode ID. */
+std::map<std::tuple<int, int, double>, RRMode> modeIdMatch;
 
 DisplayResX::DisplayResX(void)
 {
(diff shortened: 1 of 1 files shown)


phippodoplis:
Just found this and it is working for me. I had to make my monitor the primary as by default the primary is a disconnected DVI output but that's just a one-liner in .xsession. I noticed that the ticket #13100 was marked as abandoned for some reason. Is it possible to get this revived? I can test any changes you make on Debian Stretch and MythTV 29-fixes.

bennettpeter:
There is a comment in the ticket that it breaks OpenGL, with no resolution identified. If you think that the patch is good, you can re-open the ticket and let us know what you had to do and what the results were. If it fixes a problem for a few people but breaks the system for others it is not acceptable.

halovanic:
I was never able to reproduce the result of it breaking OpenGL that was reported, and without any other testers or info to go on, I gave up on this.

If there's renewed interest in this, @phippodoplis, can you see if this works under your case for OpenGL playback?  Another data point would be helpful.  I'm especially interested in which resolutions you are able to switch between.

I know this still has the primary monitor choice issue and it would need further work to handle scenarios like going from smaller resolutions to larger resolutions and altering the screen size to match the display size.

mark-kendall:
Thanks for the code. Portions of it have been committed in 962525f37f621b301cf121c3.

OpenGL playback is working (though the OpenGL code has changed completely). The missing link in associating an XRandR output with the screen in use was using QScreen::name. This should work correctly in all cases (well - I can't currently imagine a scenario where it wouldn't work).

PR #148, opened 2017-09-01
Rearrange cached files and add aging.

linuxdude42:
Create a couple of additional caching directories, and implement aging on them.  By default anything not accessed in one week is removed, and this aging is performed with mythfrontend starts and exits. (Specifically its when the MythUIHelper object is created/destroyed.) The cache directories used by this commit are all under ~/.cache and are:

1. remotecache : Existing. Copies of video screenshots from the backend. Now aged out.
2. themecache/\<themename\>-\<size\> : Existing. Now contains only theme related thumbnails. No aging.
3. thumbnails : New. Non theme related thumbnails are now stored here instead of in the themecache directories. Aged out.
4. themechooser : Location where the downloaded themes.zip is unpacked by the theme chooser. Moved here from \<themename\>-\<size\>/themechooser so now there is only ever one instance.

Fixes https://code.mythtv.org/trac/ticket/12274

### DIFF
--- a/mythtv/libs/libmythbase/mythdirs.cpp
+++ b/mythtv/libs/libmythbase/mythdirs.cpp
@@ -22,6 +22,10 @@ static QString themedir = QString::null;
 static QString pluginsdir = QString::null;
 static QString translationsdir = QString::null;
 static QString filtersdir = QString::null;
+static QString cachedir = QString::null;
+static QString remotecachedir = QString::null;
+static QString themebasecachedir = QString::null;
+static QString thumbnaildir = QString::null;
 
 void InitializeMythDirs(void)
 {
@@ -179,6 +183,10 @@ void InitializeMythDirs(void)
 
     if (confdir.length() == 0)
         confdir = QDir::homePath() + "/.mythtv";
+    cachedir = confdir + "/cache";
+    remotecachedir = cachedir + "/remotecache";
+    themebasecachedir = cachedir + "/themecache";
+    thumbnaildir = cachedir + "/thumbnails";
 
 #if defined(Q_OS_ANDROID)
     themedir        = sharedir + "themes/";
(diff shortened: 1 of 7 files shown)


bennettpeter:
I have a concern about the themechooser part. Does this mean when you switch themes it will always have to download the new theme? There may be people who like to switch themes back and forth between two themes and this may make that cumbersome. While testing I do switch themes quite often.

linuxdude42:
On Fri, 2017-09-01 at 13:59 +0000, Peter Bennett wrote:
> I have a concern about the themechooser part. Does this mean when you
> switch themes it will always have to download the new theme? There
> may be people who like to switch themes back and forth between two
> themes and this may make that cumbersome. While testing I do switch
> themes quite often.

No, my change shouldn't affect the installed themes at all.  Here's the
current behavior when you invoke the theme chooser:

1) If <confdir>/tmp/themes.zip is present and less than an
   hour old, go to 3.
2) Download themes.zip to <confdir>/tmp/themes.zip
3) Unzip to <confdir>/cache/themecache/<currenttheme>/themechooser
4) Present the theme chooser
5) If the chosen theme isn't present in /usr/share/mythtv/themes
   then downloaded it to <confdir>/themes
6) Delete all but the most recent two cached theme directories.  This
   purges things in <confdir>/cache/themecache/, not
   <confdir>/theme/.

My change only affects step 3.  Instead of unzipping themes.zip to
<confdir>/cache/themecache/<currenttheme>/themechooser in step 3 it
always unzips it to <confdir>/cache/themechooser.

As you change themes in your testing you should see the existing
behavior in step 6. Myth will delete the cached (i.e. pre-scaled)
versions of the themes from <confdir>/cache/themecache/, but it 
will never delete the full size version of the theme
from <confdir>/theme/.

David

PR #153, opened 2017-09-17
Update log_query to handle bytearray queries

tomhughes:
Recent versions of MySQLdb sometimes pass the query as a bytearray that has been encoded using the database encoding which then causes an exception as you can't pass a bytearray to join.


tomhughes:
Opened https://code.mythtv.org/trac/ticket/13123 to track this.

PR #158, opened 2017-11-05
Restore realtime priority capability for vsync thread

dlarrick:
MythTV has a settings item for using realtime scheduling for the video sync thread, but it's no longer hooked up to anything in the code. This patch restores realtime scheduling for this thread.

### DIFF
--- a/mythtv/libs/libmythtv/mythplayer.cpp
+++ b/mythtv/libs/libmythtv/mythplayer.cpp
@@ -54,8 +54,8 @@ using namespace std;
 #include "ringbuffer.h"                 // for RingBuffer, etc
 #include "tv_actions.h"                 // for ACTION_BIGJUMPFWD, etc
 
-extern "C" {
 #include "vsync.h"
+extern "C" {
 #include "libavcodec/avcodec.h"
 }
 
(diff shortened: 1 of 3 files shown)


mythtv/libs/libmythtv/mythplayer.cpp
```diff
@@ -2201,12 +2215,15 @@ bool MythPlayer::PrebufferEnoughFrames(int min_buffers)
     if (!videoOutput)
         return false;
 
-    if (!(min_buffers ? (videoOutput->ValidVideoFrames() >= min_buffers) :
```
stuarta: This code change looks unrelated to the problem it is fixing.

dlarrick: Change here is adding the log statement below. Hitting the "SetBuffering" call is a primary indicator of the problem I'm addressing with these changes. But yeah, it can go away if preferred.

stuarta: Ah right, that makes sense. It can stay :)

bennettpeter:
I have requested information in the MythTV trac ticket. Please follow up on that.

bennettpeter:
I have committed this change. I removed the extra logging. Please close  the pull request.

dlarrick:
Thanks Peter!

PR #159, opened 2017-11-05
Default to more RPi video buffs, falling back on alloc failure

dlarrick:
Experiments on my system show that 11 buffers is insufficient to avoid buffering on some US OTA HDTV content.

https://code.mythtv.org/trac/ticket/13171#ticket


bennettpeter:
I have requested information in the MythTV trac ticket. Please follow up on that.

dlarrick:
Aha! I've figured this out. It's memory related. If the Pi is memory starved and does light swapping, 4 more video buffers can make the difference between a glitch and floating through the event. I agree, then, that this change is just treating a symptom.

PR #164, opened 2018-03-28
mythtranscode: Fix #244 by detecting write errors

### LINKED ISSUE #244: mythtranscode: full disk truncates recording
Also reported at https://code.mythtv.org/trac/ticket/12602 and fixed by pull request #164.

**DATA LOSS BUG**: Disk full during lossless transcode permanently loses the end of the recording.

* **Platform**:  Ubuntu 18.04 LTS: Linux mythtv 4.15.0-118-generic #119-Ubuntu SMP Tue Sep 8 12:30:01 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

* **MythTV version**: fixes/29 but also in master

* **Package version**: mythtv/bionic,bionic,now 2:29.1+fixes.20180414.329c235-0ubuntu3 all [installed]

* **Component**: mythtranscode

Backup your test recording first! Or use a worthless recording.

Fill your disk so that free space exists but is less than the size of the large test recording you are about to transcode. Now queue a lossless transcode to permanently remove a cutlist. When disk gets completely full during the transcode, this error is not detected, and the rename of .tmp file is performed anyway, overwriting the original with the failed file. Of course this permanently loses the end of the recording. I see no evidence of the full disk or related error in mythtranscode.log, but it does clearly log the final rename.

Every time disk fills while transcode is in progress.  So in practice, only for large recording and with disk free space < recording size.

Expect the transcode to fail and original recording be left unmodified.  This lets user rerun the transcode later when he has freed more disk space.

### What do you see instead?

Incomplete transcode is renamed over the original recording, permanently losing the tail of the recording.

### Additional information

This bug is resolved by pull request #164 which detects any write/close failure.  This causes mythtv to drop the .tmp file and keep the original file unmodified.

twitham1:
Fix for #244 which was originally https://code.mythtv.org/trac/ticket/12602

Explanation of the patch:

* multiplex.h:
  the mx multiplex structure gets a new "int error", the error count.
  write_out_packs and finish_mpg now return int, the error count (so 0 = success)
  currently unused: callers test the error count instead
* multiplex.c:
  if write() fails, increment error count
  log only first 10 failures so we don't flood syslog and cause it to drop messages
  if close() fails, increment error count
* mpeg2fix.cpp:
  copy error count to static variable and pass it to pthread_exit
  capture error count in pthread_join and log/fail if non-zero

This all causes mythtv to keep the original recording if disk fills during transcode, instead of truncating the recording.  Of course user will need to go free space larger than this recording for a successful transcode.

In mpeg2fix.cpp there is also a commented section where I tried to exit on first write failure. This just deadlocked the other thread. It might be possible to cancel the other thread if you wish to abort at first write error but I was unsure how to do this. As-is, we just continue to fail and return the failure count when the reader is done. I monitored with top and the size doesn't grow so I assume the data is just being dropped rather than filling memory, which is a good thing.

### DIFF
--- a/mythtv/programs/mythtranscode/external/replex/multiplex.cpp
+++ b/mythtv/programs/mythtranscode/external/replex/multiplex.cpp
@@ -301,7 +301,7 @@ static void writeout_video(multiplex_t *mx)
 	if (write(mx->fd_out, outbuf.data(), written) != written) {
 	  mx->error++;
 	  if (mx->error <= 0) mx->error = 1; // avoid int rollover to zero
-	  if (mx->error < 10) // bug12602: log only first few failures
+	  if (mx->error < 10) // mythtv#244: log only first few failures
 	    LOG(VB_GENERAL, LOG_ERR,
 		QString("%1 writes failed: %2")
 		.arg(mx->error).arg(strerror(errno)));
@@ -651,7 +651,7 @@ int finish_mpg(multiplex_t *mx)
 		write(mx->fd_out, mpeg_end.data(), 4);
 
 	if (close(mx->fd_out) < 0) {
-	  mx->error++;	    // bug12602: close could fail on full disk
+	  mx->error++;	    // mythtv#244: close could fail on full disk
 	  if (mx->error <= 0) mx->error = 1; // avoid int rollover to zero
 	  LOG(VB_GENERAL, LOG_ERR,
 	      QString("close failed: %1")
@@ -703,7 +703,7 @@ void init_multiplex( multiplex_t *mx, sequence_t *seq_head,
 	int i = 0;
 	uint32_t data_rate = 0;
 
-	mx->error = 0; // bug12602: added to catch full disk write failures
+	mx->error = 0; // mythtv#244: added to catch full disk write failures
 	mx->fill_buffers = fill_buffers;
 	mx->video_delay = video_delay;
 	mx->audio_delay = audio_delay;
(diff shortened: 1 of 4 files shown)


twitham1:
See also #163 for original pull request against fixes/0.27.

twitham1:
So we applied all fixes Beirdo recommended here and in #163 .  Of course it still works at avoiding data loss at full disk during transcode, thereby resolving bug #12602.  So not sure what more should be done here other than merge into master.  Unless someone else wants to confirm the fix.  How do do that is described in the bug intro at https://code.mythtv.org/trac/ticket/12602

twitham1:
Hmmm, I read a recommendation to rebase against master.  But that just merged hundreds of master changes into here, hiding my actual bug fix commits somewhere in the long list.  Not sure if there is a way to undo a rebase to have a shorter commit list.  Anyway, at least the diff is still the same: I have changed only 4 files in mythtranscode only.

twitham1:
I reset my branch to current master and reduced the commit complexity down to the actual changes only.  Detailed commit history is over in #163 if needed.

twitham1:
Since @linuxdude42 most recently twiddled code in mythtranscode, I wonder if he would be willing to take a look at this?

linuxdude42:
The changes look reasonable to me. I see only one minor flaw, which is that the "|| 0" in "return mx->error || 0;" is redundant.

twitham1:
Good point, that || 0 is a habit from my Perl coding, for upgrading an undef to zero.  But in this C code the error is already initialized to 0 so this is not needed.  You may remove my silly but harmless mistake if you like.

So glad to see this bug fixed, thank you sir!

PR #165, opened 2018-05-13
Allow option to disable qtwebkit to compile with QT From Homebrew 

mmisiewicz:
This change permits the disabling of `qtwebkit` to allow MythTV to be built with QT from Homebrew.


mark-kendall:
QTWebkit is no longer required to build master

PR #166, opened 2018-07-30
python: Handle massively negative initial zoneinfo entries.

ijc:
Some zoneinfo files contain such "big bang hack" entries as a workaround for a
bug, this was recently reverted upstream but such files are still around in the
wild (e.g. Ubuntu 14.04, current Debian testing). Some info on this can be
found in:

https://github.com/eggert/tz/commit/83c119f4d5d48ba37d73e42b0f12da3ae06b6c3f

With this a:

    from MythTV import datetime
    import time
    tnow = time.time()
    print datetime.fromtimestamp(tnow).strftime("%Y%m%d%H%M%S %z")

Now produces the correct time. To reproduce I needed to run with
`TZ=Europe/Vienna` since I am in GMT/BST. Prior to this fix when running in
during Northern hemisphere daylight savings time (i.e. today) the code would
produce timezone of +0100 instead of the correct +0200. Similarly with
`TZ=Americ/Los_Angeles` an incorrect -0800 becomes -0700 with the fix.

The change only ignores initial entries of the bad for, since it would be
unexpected to see them once we have seen a transition for the "modern era"
(which I define as "something time.gmtime is happy with").

### DIFF
--- a/mythtv/bindings/python/MythTV/utility/dt.py
+++ b/mythtv/bindings/python/MythTV/utility/dt.py
@@ -158,11 +158,24 @@ class posixtzinfo( basetzinfo ):
                     1)
             return
 
-        transitions = []
-        for i in range(counts.transitions): # read in epoch time data
+        transitions = [None] * counts.transitions
+        # First transition which is in range for gmtime. Some zoneinfo
+        # files have massively negative leading entries for e.g. the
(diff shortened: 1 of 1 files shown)


ijc:
Merged as ea500ae37442de018494baf9d8f000949fb8a6d3 via the ticket.

PR #169, opened 2018-09-09
Fix python bindings not reconnecting to database after a long pause.

sir-maniac:
Fix a version check that ensures the python MySQLdb library auto
reconnects in v1.2.2 or higher (earlier versions default to
auto-reconnect).

trac ticket: https://code.mythtv.org/trac/ticket/13320


rcrdnalor:
This was fixed by commit 9127e34

PR #175, opened 2018-10-30
Compare to `None` using identity `is` operator

janisozaur:
This is a trivial change that replaces `==` operator with `is` operator, following PEP 8 guideline:

> Comparisons to singletons like None should always be done with is or is not, never the equality operators.

https://legacy.python.org/dev/peps/pep-0008/#programming-recommendations


janisozaur:
https://code.mythtv.org/trac/ticket/13343

rcrdnalor:
This was fixed on master by commit 2e7e9e8
and on v31 by commit c0b8b6e

linuxdude42:
See above. This was reported fixed on 2020-10-02.

PR #178, opened 2019-02-26
Prevent data corruption in dvbchannel.cpp causing hang

simonhyde:
m_is_open was being corrupted by multiple simultaneous modifications,
leading to it managing to create infinite loops within its underlying
data structure. This patch extends the locking of m_hw_lock to cover
all uses of m_is_open, preventing issues.


kmdewaal:
Fixes #13415

PR #186, opened 2019-11-04
initial code for porting mythlink to python

dawesc:
There are a lot of places in the documenation talking about mythlink pending transcode to python from perl, i hope to move things along by writing the initial port, i've tested this and am happy to the extent that links are being created initally and then continually. Hopefully this will provide a platform for moving this utility forward though it's been coded by translating the perl to python and not to try to increase performance or do things in a more modern approach.


dawesc:
Xref trac ticket

https://code.mythtv.org/trac/ticket/13499

PR #187, opened 2019-11-14
0.30+fixes+ime

Artanejp:
This is support of input method extension (IME) to input CJK(or another) string via input method (i.e. IBUS, fcitx...).
This adds some APIs to libmythui and using QInputMethod.
This is initiali support (and limited support).

Regards,
Ohta.

### DIFF
--- a/mythtv/libs/libmythui/mythdialogbox.cpp
+++ b/mythtv/libs/libmythui/mythdialogbox.cpp
@@ -399,6 +399,14 @@ void MythDialogBox::AddButton(const QString &title, const char *slot,
     m_buttonList->GetVisibleCount();
 }
 
+bool MythDialogBox::inputMethodEvent(QInputMethodEvent *event)
+{
+    if (GetFocusWidget()->inputMethodEvent(event))
+        return true;
+
+	return false;
+}
+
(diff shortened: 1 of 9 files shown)


bennettpeter:
Why did you change the base branch from master to fixes/30?

Artanejp:
I am in Japan, MythTV must apply ISDB-JP patches for 30/fixes.

So, I can't test based on master.

So, I changed base from master to 30/fixes.

Sorry.

P.S:

I checked building IME modifies with master,  need with some changes, with none FTBFS (but don't be able to test working for me).

See https://github.com/Artanejp/mythtv/tree/0.30+ime-extension .

linuxdude42:
This seems like a reasonable thing to go into the code, but I'm not familiar with IME. How would I test this?

linuxdude42:
I just committed these changes to the master branch. They will be in the upcoming v32 release. Are your ISDB-JP patches something that could also be committed to the main repository?

Artanejp:
Sorry for later, and thanks to commit these changes.
ISDB-JP patches are based by some anonymous of 2ch.net (or 5ch.net) at  first half of 2010s.
So, some of (or all of) them needs to re-make.
Thanks for your kindness.

PR #189, opened 2019-11-20
Reduce memory usage of mythfilldatabase by a factor 6, speed up by 15%

dingo35:
Mythfilldatabase uses QDomDocument to parse and store the XML data that is read; according to QDomDocument documentation this object is not meant to handle large XML files; QXmlStreamReader should be used in these situations.

This pull request replaces QDomDocument by QXmlStreamreader; loading a 222 MB XML file used 2,5GB on memory, after these changes it only uses 420MB, and speeds up the loading process by approx. 15% .


dingo35:
Committed https://github.com/MythTV/mythtv/commit/a9aa006139da24cbad861a2c25d57fa3f71aba2a

PR #190, opened 2020-01-04
MythDisplayX11: Use accurate XRandR refresh rates

Glidos:
This fixes automatic frame-rate switching when using modern nvidia drivers.

Reapply parts of Mark Kendall's use of the more modern XRandR API from the
render branch, specifically the changes to GetVideoModes and
SwitchToVideoModes.

Also take on part of a later commit to filter out interlaced modes.

GetVideoModes requires selection of an output matching the screen currently used by mythfrontend.
In the original version of this code from the render branch, a reference to the screen was held in a
member variable of the class. For the fixes/29 branch, the screen has to be derived differently. The
calls to do so were also suggested by Mark.

### DIFF
--- a/mythtv/libs/libmythui/DisplayRes.cpp
+++ b/mythtv/libs/libmythui/DisplayRes.cpp
@@ -236,7 +236,7 @@ bool DisplayRes::SwitchToCustomGUI(int width, int height, short rate)
 }
 
 const std::vector<double> DisplayRes::GetRefreshRates(int width,
-        int height) const
+        int height)
 {
     double tr;
     std::vector<double> empty;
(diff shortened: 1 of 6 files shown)


mark-kendall:
Paul - do you need me to apply this?

I think it needs a small update for 5cdda63eaf2bd5fbddceac27cf08a5295bf28c7b and probably 1d4d51b0d9523407f0d4f201822bf6bb03e4c1c8

Glidos:
I'm not sure how to go about doing that. I looked at 5cdda63eaf2bd5fbddceac27cf08a5295bf28c7b and there are seemingly important uses and changes to UpdateCurrentMode, which doesn't exist at the point in history where I'm working. I think UpdateCurrentMode was introduced in d6798be0c5adc0c40dd3f7f9e94eab5a4de96613, but that looks very difficult to reapply.

mark-kendall:
Paul - are you still using 0.29? Is this still relevant? As you say, I think the changes are non-trivial but I cannot test 0.29

mark-kendall:
@Glidos Closing this as I don't think it is going to progress...

PR #191, opened 2020-02-26
mythmusic fixes for multiple discs

ijc:
I noticed that mythmusic wasn't handling multi-disc albums correctly.

* the flacvorbis metadata handler didn't know about the `DISCNUMBER` tag so I added it (I added `TOTALTRACKS` and `TOTALDISCS` while I was there too). My goto reference for tag names is https://picard.musicbrainz.org/docs/mappings/ (all my music is tagged using Picard so this is what I've ended up with).
* once the discnumber was in the DB I noticed that the album sort wasn't handling it correctly, e.g. for a two disc album it would sort disc.track-wise as 1.1, 2.1, 1.2, 2.2, etc rather than 1.1, 1.2, ..., 2.1, 2.2. Incorporate the disc number into the album order to correct this. I believe a single CD can have at most 99 tracks so use a multiplier of 100. This means a 10 disc box set would clash with album map order offset of 1000, so bump that to 10000.

I've been running this on my v30 based production system for a week and have tested against master.

### DIFF
--- a/mythplugins/mythmusic/mythmusic/playlist.cpp
+++ b/mythplugins/mythmusic/mythmusic/playlist.cpp
@@ -386,8 +386,10 @@ void Playlist::shuffleTracks(MusicPlayer::ShuffleMode shuffleMode)
                     }
                     else
                     {
-                        album_order = Ialbum->second * 1000;
+                        album_order = Ialbum->second * 10000;
                     }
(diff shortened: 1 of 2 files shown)


ijc:
Xref to trac: https://code.mythtv.org/trac/ticket/13592

PR #193, opened 2020-03-28
honour port setting in mythconverg backup/restore perl scripts

metaron-uk:
While trying to move my mythbackend sql database to my syndology nas (which uses port 3307 for its mariadb server) I discovered that mythconverg_restore.pl and mythconverg_backup.pl don't honour the port setting in config.xml when trying to login to the mysql server.

Confirmed working using mythtv/31 on my pi4 gentoo based mythtv installation


kmdewaal:
@rcrdnalor this PR looks OK to me, what do you think? Shall we merge it?

PR #197, opened 2020-04-13
libmythui: Hide the mouse cursor on gnome-shell under xwayland

delxg:
Prior to this change the mouse cursor would be visible over the UI upon startup. It would remain visible despite pressing keys or starting a video. The only way to make it disappear was to plug in a mouse and move it.

I've included this patch in the [Arch Linux mythtv-git](https://aur.archlinux.org/packages/mythtv-git) package. Thanks for a taking a look.

Sorry for missing a Trac ticket. I read through everything on the [TicketHowTo](https://code.mythtv.org/trac/wiki/TicketHowTo) but I couldn't figure out how to create a new Trac account. Without an account I cannot login. Visiting the [new ticket](https://code.mythtv.org/trac/newticket) page just tells that `TICKET_CREATE privileges are required`.

### DIFF
--- a/mythtv/libs/libmythui/mythmainwindow.cpp
+++ b/mythtv/libs/libmythui/mythmainwindow.cpp
@@ -1004,12 +1004,6 @@ void MythMainWindow::Init(bool mayReInit)
     // screen position ends up stuck. Adding a border temporarily prevents this
     setWindowFlags(windowFlags() & ~Qt::FramelessWindowHint);
 
-    if (!inwindow)
-    {
-        LOG(VB_GENERAL, LOG_INFO, "Using Frameless Window");
-        flags |= Qt::FramelessWindowHint;
-    }
-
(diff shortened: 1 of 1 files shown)


ijc:
>  I couldn't figure out how to create a new Trac account.

It's a bit unobvious but on the login page on the RHS of the box (the one with the username/password entry on the LHS) there is a github logo -- if you click that you should be able to login with github.

delxg:
I did try that and got an error from `https://sso.mythtv.org/auth/realms/MythTV/login-actions/first-broker-login` saying:

> We are sorry...
> Invalid username or password."

mark-kendall:
To be honest - I'm struggling to see why this is appropriate. The proposed change will potentially break full screen/borderless display for many systems; and doesn't seem to directly address the reported problem.

I suspect it is just a coincidence that this fixes the mouse problem in this case - a closer inspection of the Qt source might suggest why.

PR #199, opened 2020-07-07
extend metadatagrabber timeout to 3 minutes

bhulsken:
ttvdb and tmdb are sometimes quite slow to respond to queries, and the 60 seconds timeout is really too short. Extending it to 180s (3 minutes) solves issues for me, and allow fetching of metadata for all movies / series in my library (which are many).

I have a very fast (fiber) internet connection, so the issue is not my connection, also multiple reports on this issue can be found on the web. For some reason, slower replies of tmdb seem linked to certain movies. Meaning that some queries yield a reply in less than 60 seconds, yet consistent repeated queries for certain movies never return a result within 60 seconds.


warpme:
I have the same observations and confirm this change reduces false positives where user thinks issue is within mythtv metagrabber while real root cause is slow repose from metadata source.

PR #269, opened 2020-10-21
changed exit prompted to only appear in a single case

majurgens:
changed exit prompted to only appear in a single case - where the backend and frontend are on the same system

any other prompting is excessive as there is already an unlikely-to-be-an-accident combination of keystrokes to get to this point
even if the user did mistakenly press the unlikely-to-be-a-mistake key sequence to shutdown, then this is unlikely to be of significant consequence to the user as they can just restart the computer/application

##### Checklist

- [ x] code follows the [MythTV Coding Standards](https://www.mythtv.org/wiki/Coding_Standards)
- [ x] documentation added/updated/removed where necessary

### DIFF
--- a/mythtv/programs/mythfrontend/exitprompt.cpp
+++ b/mythtv/programs/mythfrontend/exitprompt.cpp
@@ -263,12 +263,15 @@ void ExitPrompter::Confirm(MythPower::Feature Action) const
     else 
     {
       // no prompts required, take a specific action required
+      // calling these exitprompter functions with (true) fails build
+      bool isTrue;
+      isTrue=true;
       if (Action == MythPower::FeatureShutdown)
-          ExitPrompter::DoHalt(true);
+          ExitPrompter::DoHalt(isTrue);
       else if (Action == MythPower::FeatureRestart)
-          ExitPrompter::DoReboot(true);
+          ExitPrompter::DoReboot(isTrue);
       else if (Action == MythPower::FeatureSuspend)
-          ExitPrompter::DoSuspend(true);
+          ExitPrompter::DoSuspend(isTrue);
     }
 
     gContext->SetDisableEventPopup(false);
--- a/mythtv/programs/mythfrontend/exitprompt.cpp
+++ b/mythtv/programs/mythfrontend/exitprompt.cpp
@@ -36,7 +36,7 @@ void ExitPrompter::ConfirmHalt() const
     Confirm(MythPower::FeatureShutdown);
 }
 
-void ExitPrompter::DoHalt(const bool Confirmed) const
+void ExitPrompter::DoHalt(const bool Confirmed)
 {
     if (!Confirmed)
         return;
@@ -67,7 +67,7 @@ void ExitPrompter::ConfirmReboot() const
     Confirm(MythPower::FeatureRestart);
 }
 
-void ExitPrompter::DoReboot(const bool Confirmed) const
+void ExitPrompter::DoReboot(const bool Confirmed)
 {
     if (!Confirmed)
         return;
@@ -97,7 +97,7 @@ void ExitPrompter::ConfirmSuspend(void) const
     Confirm(MythPower::FeatureSuspend);
 }
 
-void ExitPrompter::DoSuspend(const bool Confirmed) const
+void ExitPrompter::DoSuspend(const bool Confirmed)
 {
     if (!Confirmed)
         return;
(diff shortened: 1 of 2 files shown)


bennettpeter:
There is a setup option on the frontend  (Setup->General->shutdown/reboot settings) where you can choose the actions to be performed. I believe there used to be an option for just shutdown with no prompt. I don't know what happened to that, but I would prefer an option there, so that you can select whether or not to be prompted.

For me, the prompt is essential. When I am several menus deep in the frontend and I want to get to the main menu, I press "exit" many times, and pressing one time too many would exit the frontend, which is painful since it takes a long time to restart again if in fact I did not want to exit the frontend.

gigem:
The option to prompt or not is controlled using key bindings.  In the
Main Menu context, bind keys to either the EXIT or EXITPROMPT actions.

David
-- 
David Engel
david@istwok.net

majurgens:
Sorry, let me be more explicit.

The standard workflow for shutdown/reboot/exit (when these options are configured) is (there may be some variants depending on configuration):

1. Press ESC
2. Get "Shutdown/Reboot/Exit" prompt (https://www.edcint.co.nz/tmp/myth_standard_prompt.jpg)
3. Press Up-Arrow
4. Press Enter
5. Get extra prompt "Are you sure you want to .... OK/Cancel" (https://www.edcint.co.nz/tmp/myth_extra_prompt.jpg)
6. Press Up-Arrow
7. Press Enter

This pull request is referring to the prompt at Step 5. 

@gigem you are referring to the action that occurs when you press ESC at the mainmenu. By changing the bind keys you either get the option to not prompt (EXIT action) or to prompt by the dialog at Step 2 (EXITPROMPT action).
@bennettpeter you are referring to configuration of the prompt at Step 2 and the options that it shows 

I am not proposing to change any of the prompt at Step 2, not keystrokes to activate options from it, so users can still chose to be prompted or not at Step 2.

It is the prompt at Step 5 that extra and unnecessary since in order to get to Step 5, the user has already performed a-highly-likely-to-be-intentional-key-combination to shutdown. Getting to Step 4 is unlikely to be an accident and does not require additional prompting.

gigem:
@majurgens.  Ah, okay.  I remove the ESC binding from the exit actions and use another, uncommon key to exit unconditionally.  I suggest you do the same as it seems Peter and others like the extra protection.

majurgens:
> 
> 
> @majurgens. Ah, okay. I remove the ESC binding from the exit actions and use another, uncommon key to exit unconditionally. I suggest you do the same as it seems Peter and others like the extra protection.

Yep, change the key binding from ESC to whatever other key you like - this is still only referring to Step 1 of the workflow I describe above and this pull request is not changing that. By changing the key binding all you have done is change the way you enter the workflow I describe.

This pull request, if implemented, would still not have what you refer to as unconditional exit - you still, by default, are prompted for exit confirmation at Step 2.

I am trying to remove the confirmation of the confirmation at Step 5

gigem:
Now you are misunderstanding.  By properly configuring your keys and options, you will not get prompted when exiting mythfrontend.  I don't get prompted and haven't in several years.  I suggest again that you consider that option.  The reason is it's seems unlikely that your pull request will be accepted.  I'm ambivalent but won't pull it unilaterally, Peter has indicated he doesn't want it and none of the other devs have expressed interest in it.

majurgens:
I'm not trying to "not get prompted", I'm trying to not get prompted twice. At the moment I get prompted twice.
This double prompt never happened on v30, it was introduced in v31. 
Maybe I have messed it all up. Perhaps you can tell me how I can configure the following flow:

1. Press ESC
2. Get "Shutdown/Reboot/Exit" prompt (https://www.edcint.co.nz/tmp/myth_standard_prompt.jpg)
3. Press Up-Arrow
4. Press Enter
5. Machine shuts down

One thing that I have not mentioned is that this only happens for a remote frontend.

mark-kendall:
I very deliberately added the extra confirmation after testing support for the new power management code that went into 0.31.

- I found it far too easy to accidentally power off my system with some fast/inaccurate keypresses
- recovering from an accidental power event is, imho, not a trivial task. It takes me at least 2-3 minutes and requires me to actually get off my backside and press a physical button on the box (and bear in mind that there will be many users for whom physical access to their box is limited)
- the code cannot account for other (non mythtv) processes that are running on the system - which may be running intermittently
- one of the original aims of new code was to add a 'cancel' option to the power event (with a configurable timeout) to guard against user 'errors'. Unfortunately this doesn't work as intended across all supported platforms as once the power event is triggered we lose the UI (this may be fixable).

That said, I will knock up a patch that adds an extra setting in the Shutdown/Reboot settings screen to make the additional confirmation prompt configurable (defaulting to 'prompt') - does that sound like a reasonable compromise?

majurgens:
The extra setting would be great and allow users the flexibility.

PR #303, opened 2021-01-18
Add optional metadata grabber for television: tvmaze

rcrdnalor:
See
https://www.tvmaze.com/
https://www.tvmaze.com/api

This commit adds an alternative metadata grabber for TV-Series.
The tvmaze metadata grabber follows the indentions in
https://www.mythtv.org/wiki/MythTV_Universal_Metadata_Format

To enable this grabber, run in mythtv frontend
"Setup" -> "Artwork and Data Sources"
and select 'TVMaze.com' for the default TelevisionGrabber.

Once done, you can check this global setting with mysql:

[mythconverg]> select * from settings where value like '%grabber%';
+--------------------+-------------------------------+----------+
| value              | data                          | hostname |
+--------------------+-------------------------------+----------+
| TelevisionGrabber  | metadata/Television/tvmaze.py | NULL     |
| MovieGrabber       | metadata/Movie/tmdb3.py       | NULL     |
+--------------------+-------------------------------+----------+

Any further metadata update will then use these grabbers.

Note: The tvmaze grabber is compatible to python 2.7 and 3.6+.

##### Checklist

- [x ] documentation added/updated/removed where necessary

Attributes of this  tvmaze grabber
pros:
- free of charge
- compatible to mythtv/fixes/v29+

cons:
- only English, no localization
- less data


bennettpeter:
Python is totally new to me - 
I see you added a comment 'added python2 compatibility'. I had asked Mark Spieth about it and he said not to worry, we need not support python 2. So I am ignoring that in my changes to tmdb3.
Does the license from Lachlan Mackenzie allow the code to be included in a GPL2 project?

rcrdnalor:
The licence from Lachlan Mackenzie is a MIT licence.
AFAIK, this means, that one can re-use the code under GPL2, once you mentioned the licence and the copyright statement and add a statement, what was changed in the code. This is what I've done. 
Suggestions/Improvements/Corrections are always welcome. 

The answer from M. Spieth in respect to python3 is correct if we use a stand-alone grabber: python3 is supported on all our OS-targets.
If we try to re-use existing python libraries, like the tmdb3-api, which is part of the MythTV's python bindings, 
*and* we want to be compatibel to fixes/29+ , then the TeleVision grabber script should understand python2 as well.
This will give us an easy upgrade path to all users of older MythTV versions.

I have not realized that themoviedb.org offers a good database for tv-series as well, until you mentioned it.
I did a look at the existing tmdb3-api, and it looks promising to add a TeleVision grabber using that API.
This would give us an easy path to support MythTv v29+.

I have a test-setup that supports python 3.6+ and MythTV from v.29 to master.
If you don't mind, I can take over the tmdb3 television grabber part from you 
and I will create a pull request for it.
Then we can compare the performance between these new grabbers.
I personally only want to support one them ;-)

bennettpeter:
There is a lot missing from tmdb3 for TV shows. It looks like somebody started on it but never finished.
I am making some progress on the tmdb3 grabber, learning a lot about python in the process. I would like to finish what I am doing, and you can review it for me.

bennettpeter:
I tested this with a few shows, both in recordings and in videos. All is working fine. I have not looked at the code.

bennettpeter:
Roland, I recommend you merge this request.

rcrdnalor:
See
https://www.tvmaze.com/
https://www.tvmaze.com/api

This commit adds an alternative metadata grabber for TV-Series.
The tvmaze metadata grabber follows the defintions in
https://www.mythtv.org/wiki/MythTV_Universal_Metadata_Format

To enable this grabber, run in mythtv frontend
"Setup" -> "Artwork and Data Sources"
and select 'TVMaze.com' for the default TelevisionGrabber.

Once done, you can check this global setting with mysql:

[mythconverg]> select * from settings where value like '%grabber%';
+--------------------+-------------------------------+----------+
| value              | data                          | hostname |
+--------------------+-------------------------------+----------+
| TelevisionGrabber  | metadata/Television/tvmaze.py | NULL     |
| MovieGrabber       | metadata/Movie/tmdb3.py       | NULL     |
+--------------------+-------------------------------+----------+

Any further metadata update will then use these grabbers.

Note: The tvmaze grabber is compatible to python 2.7 and 3.6+.

bennettpeter:
I notice that the urls in endpoints.py have https on some and http on others. Any reason for the inconsistency? 

Some people with very old versions of python are having certificate errors, so it may be better to revert to http if that will work.

rcrdnalor:
I will look into this.
Thanks for reporting.

Gribnif:
Please be careful with defaulting to http:// URLs. It's been my experience that, with browsers now actively complaining when a web site dos not use SSL, some API maintainers have silently started forcing the use of https://URLs, thus breaking old clients. You might check with the sites in question to verify that they intend to continue to support non-SSL.

stuarta:
It is bad security practice to use authentication with any site not using https. We should aim to be using https for all endpoints.

ijc:
Exposing all the users who have kept up to date with the supported versions to potentially malicious responses via MITM attacks for the sake of one report from a user on an older version (which doesn't seem to have been definitively diagnosed as a cert issue so far) doesn't seem like the right trade-off -- old sayings about baby and bath water come to mind :grin:.

SteveErl:
I've been experimenting with the new tvmaze.py script in my MythTv (v31) system. In general it works, but so far every test  fails to return "fanart" or "banner" URLs. I thought maybe the TvMaze database doesn't have those graphic elements, but after checking, that is not the case. For example,
https://www.tvmaze.com/shows/36947/resident-alien/gallery
shows 3 **_posters_** ('coverart' to MythTv), 1 **_banner_**, and 2 **_backgrounds_** ('fanart' to MythTv)

```
/usr/share/mythtv/metadata/Television/tvmaze.py  -l en -a US -M "Resident Alien"
<metadata>
  <item>
    <title>Resident Alien</title>
    <description>Resident Alien is a dark, twisted and comedic fish-out-of-water story that follows a crash-landed alien named Harry who, after taking on the identity of a small-town Colorado doctor, slowly begins to wrestle with the moral dilemma of his secret mission on Earth — ultimately asking the question, "Are human beings worth saving?"</description>
    <inetref>36947</inetref>
    <collectionref>36947</collectionref>
    <language>en</language>
    <releasedate>2021-01-27</releasedate>
    <year>2021</year>
    <images>
      <image type="coverart" url="https://static.tvmaze.com/uploads/images/original_untouched/288/722418.jpg" thumb="https://static.tvmaze.com/uploads/images/medium_portrait/288/722418.jpg"/>
    </images>
  </item>
</metadata>
```
Personally, the "fanart" background image is my favorite type of these images, but it seems like all three image types should be retrievable, when the database has them populated.

Could you translate TvMaze **background** URLs into "fanart", and add in the **banner** URLs to the output of tvmaze.py? Thanks for developing this tool!

SteveErl:
I've implemented a fix for the missing **fanart** and **banner** issue I mentioned above. #336

PR #304, opened 2021-01-22
Modify tmdb3 grabber to support TV in addition to Movies

bennettpeter:
The tmdb3 grabber is modified so that you can pass an extra parameter
"-T television" to look up TV shows. Since MythTV does not support extra
parameters on the command line, there is a tiny frontend program
Television/tm3dbtv.py that MythTV recognizes and that invokes the code
from tm3db.py with the "-T television" option.

To enable this grabber, run in mythtv frontend
"Setup" -> "Artwork and Data Sources"
and select "TheMovieDB.org V3 television" for the default Television
Grabber.

##### Checklist

### DIFF
--- a/mythtv/programs/scripts/metadata/Television/tmdb3tv.py
+++ b/mythtv/programs/scripts/metadata/Television/tmdb3tv.py
@@ -1,16 +1,14 @@
 #!/usr/bin/env python
 # -*- coding: UTF-8 -*-
 # ----------------------
-# Name: tmdb3.py
+# Name: tmdb3tv.py
 # Python Script
 # Author: Peter Bennett
 # Purpose:
 #   Frontend for the tmdb3 lookup.py script that now supports both Movies and
-#   TV shows. This frontend supports Movies
+#   TV shows. This frontend supports TV shows.
 # Command example:
 # See help (-h) options
-#
-# Code that was originally here is now in tmdb3/lookup.py
 #
 # License:Creative Commons GNU GPL v2
 # (http://creativecommons.org/licenses/GPL/2.0/)
@@ -58,12 +56,16 @@ def main(showType, command):
                       dest="test", help="Perform self-test for dependencies.")
     parser.add_option('-M', "--movielist", "--list", action="store_true", default=False,
                       dest="movielist",
-                      help="Get Movies. Needs search key.")
+                      help="Get Television Series List. Needs search key.")
     parser.add_option('-D', "--moviedata", "--data", action="store_true", default=False,
-                      dest="moviedata", help="Get Movie data. " \
-                      "Needs inetref. ")
+                      dest="moviedata", help="Get TV Episode data. " \
+                      "Needs inetref, season and episode. ")
     parser.add_option('-C', "--collection", action="store_true", default=False,
-                      dest="collectiondata", help="Get Collection data.")
+                      dest="collectiondata",
+                      help='Get a television Series (collection) "series" level information')
+    parser.add_option("-N", "--numbers", action="store_true", default=False, dest="numbers",
+                        help="Get television Season and Episode numbers. " \
+                        "Needs title and subtitle or inetref and subtitle.")
     parser.add_option('-l', "--language", metavar="LANGUAGE", default=u'en',
                       dest="language", help="Specify language for filtering.")
     parser.add_option('-a', "--area", metavar="COUNTRY", default=None,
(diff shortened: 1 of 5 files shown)


bennettpeter:
This is lightly tested. I will do a lot more testing before merging it.

mspieth:
What you could do is have a soft link to tmdb.py  with a different name and check $0 for the name to determine function. This would be like busybox which is used in ram disks.

mythtv/programs/scripts/metadata/Television/tmdb3tv.py
```diff
@@ -0,0 +1,27 @@
...
+#
+__title__ = "TheMovieDB.org for TV V3"
+__author__ = "Peter Bennett"
+
+import sys, os
+filename = os.path.abspath(__file__)
+# append ../ to the path so that it finds the Movie package
+sys.path.append(os.path.dirname(os.path.dirname(filename)))
+from Movie import tmdb3
+tmdb3.main('television', 'tmdb3tv.py')
```
mspieth, line 27: set exit code here so that failure is rippled up to caller

mythtv/programs/scripts/metadata/Movie/tmdb3.py
```diff
@@ -358,22 +547,42 @@ def main():
...
+
     except RuntimeError as exc:
         sys.stdout.write('ERROR: ' + str(exc) + ' exception')
+        import traceback
+        traceback.print_exc()
         sys.exit(1)
 
 if __name__ == '__main__':
-    main()
+    main("movie",'tmdb3.py')
```
mspieth, line 588: A good construct is sys.exit(main(....)) and use return in other functions to allow for easier unit tests.

rcrdnalor (commented):
Below my remarks:
The option to search collections ("-C") is no limited to television grabber.
I recall that mythtv is calling this option for movies as well, as a last resource.
At least on older version (v30-), i have seen this on the logs.
For consistency reasons, I would prefer to implement this search for movies as well.

The tmdb3 api is designed for interactive searches as well:
See mythtv/bindings/python/tmdb3/scripts/pytmdb3.py.
This shell offers now less options/capabilities, because a lot of searches are
disabled by now in favour of speed for mythtv.
I would say, that this is not the intention of the original author of tmdb3.
To overcome this limitation, I prefer to set a (global) variable in the tmdb3_api
and based on this variable we do limited or extensive search.

The file tmdb3tv.py has now a hidden dependency to tmdb3.py, which is not visible
to developers or integrators of MythTV and can be easily overlooked in the future.
Furthermore, Python considers the manipulation of sys.path as a last option
to solve an import problem. This path manipulation is mostly used by testing
and mocking scenarios.
To solve this, I propose to create a common file located in mythtv/bindings/python
and have two separate scripts for tv and movies using this common part.

bennettpeter:
My purpose is to get tmdb3 working for TV in a basic fashion with minimum changes.

I tested the existing tmdb3.py with -C for movies. Passing -C and a valid movie inetref returns an empty xml file. Passing -C with a title that contains no space also returns an empty xml file. Passing -C with a title that includes a space throws an exception with stack trace. Fixing existing bugs with movies, if this is a bug, I see as a separate project. I can remove the check and let it revert to the former failure conditions.

Regarding the disabling of certain searches, I commented parts of season, episode, series and created some base classes that exclude loading of subsidiary sections. Without doing this a search on tv series was taking a very long time, so long that I never completed one. After waiting several minutes I canceled it. I suspect that themoviedb.org has some logic that detects if it gets a flood of requests from one location and slows down or stops responding.

Regarding the hidden dependency, I will move the main logic of tmdb3.py into the bindings as you suggest and use a small program in each of the metadata directories that calls the needed functions based on input parameters.

rcrdnalor:
An empty xml file with the '<metadata> </metadata>' tag means that nothing was found.
That is the expected behaviour. The logs will show that accordingly.
If we return absolutely nothing, the logs will raise an error condition.
Grabbing of correct metadata is one of the most discussed items in the user forum and lists.
We should be very sensible in that way.

I am not against that mythtv requests only a subset of the searches tmdb3_api is offering.
My proposal is, to limit these amount of searches only when mythtv requests data from tmdb.
In all other cases, they should be allowed, and not disabled in a hard-coded way 
(by commenting out this code).
With a global variable in tmdb3_api.py and a setter function for this variable,
one can easily guard those unwanted functions in case of mythtv is grabbing data.

rcrdnalor:
Somehow, the mentioned xml tag was cut out in previous commend:
`'<metadata> </metadata>'` was meant.

bennettpeter:
I have done more testing, and running this on my main production system. Working fine now.

bennettpeter:
This pull request has been applied directly to master as commit f53465de76

PR #336, opened 2021-03-15
Tvmaze fix

SteveErl:
Testing the tvmaze.py script revealed that it only outputs a single **coverart** path. It's missing **fanart** and **banner** URLs, even when they are available from the tvmaze.com website. In addition, it's important to output image descriptions for all of the available **coverart**, **fanart**, and **banner** artworks. When a user is not satisfied with the automatically chosen artwork, they can edit the Metadata Options for a Recording rule. In there, the options to **_Find Coverart Online_**, **_Find Fanart Online_**, and **_Find Banners Online_** are only useful if the alternative images are available to pick from.

This commit adds routines to retrieve all the artwork types, and updates tvmaze.py to emit the image descriptors for each one. 

##### Checklist
- [x] contribution does not duplicate one of our [existing pull requests]
- [x] contribution is in a branch rebased against [master]
- [x] commits are logically organised and have [good commit messages]


rcrdnalor (changes requested):
The value of 'show.info.might' retrieved from tvmaze could be
 - incomplete
 - missing
 - wrong
 
In this case, the conversion to float results in an uncaught exception (TypeError, ValueError)

rcrdnalor (commented):
it is not clear to me, why fetching additional
artwork is only done for the options "-M title" and "-C inetref".
These options are used to provide only basic data for a TV-series

rcrdnalor:
Steve,
thank you for adding additional features to the TVmaze grabber.

Below some background information about how this grabber is used 
within MythTV when fetching data for TV:

Command line options used by MythTV: 

-M title
returns a list of items related to the title, and
should contain correct title, a description, inetref and a set of 'global' artworks.

-N title subtitle 
and
-N interef subtitle
should deliver detailed information about the series episode.

-D inetref season episode
should also return detailed information of an already known episode.

-C inetref or collectionref
is the last resource, which is called to fetch data, and 
similar to '-M title' should only return 'global' data,
applicable to the whole series.

Given above information, it is not clear to me, why fetching additional
artwork is only done for the options "-M title" and "-C inetref".

We should distinguish between 'essential' data and 'other' data requested by MythTV.
For robustness reasons, fetching 'other' data should not cause any error or hinder
the script to report the 'essential' data.
Often, these 'other' data are
 - incomplete
 - missing
 - wrong
and the python script needs a lot of 'try - except' rules to handle these data.

SteveErl:
I've added exception catching blocks around the conversions of **show_info.weight** to a float. Thanks for that suggestion.

I've pulled the code which generates show-based artwork lists into its own function, _get_show_art_lists_(), which now gets called from three places. The -N and -D invocation options now generate the image lists, as was requested.

On my MythTv system, where I mostly use Manual Record rules, the mythmetadatalookup.log Info level messages show first a call using the -N option, then a call using the -M option, and finally a call using the -C option. The -N option **always** fails because the subtitle is always filled with a _timestamp_. The -M option generally works fine, and is used to find the inetref for the best match. Then the -C syntax is used using the inetref which was picked.

```
Mar 17 00:15:20 Doom mythmetadatalookup: mythmetadatalookup[19691]: I MetadataDownload metadatagrabber.cpp:430 (RunGrabber) Running Grabber: /usr/share/mythtv/metadata/Television/tvmaze.py -l en -a US -N The Daily Show with Trevor Noah 2021-03-16 22:00:00
Mar 17 00:15:21 Doom mythmetadatalookup: mythmetadatalookup[19691]: I MetadataDownload metadatagrabber.cpp:430 (RunGrabber) Running Grabber: /usr/share/mythtv/metadata/Television/tvmaze.py -l en -a US -M The Daily Show with Trevor Noah
Mar 17 00:15:23 Doom mythmetadatalookup: mythmetadatalookup[19691]: I MetadataDownload metadatacommon.cpp:1105 (ParseMetadataItem) Result Found, Season 0 Episode 0
Mar 17 00:15:23 Doom mythmetadatalookup: mythmetadatalookup[19691]: I MetadataDownload metadatagrabber.cpp:430 (RunGrabber) Running Grabber: /usr/share/mythtv/metadata/Television/tvmaze.py -l en -a US -C 3928
Mar 17 00:15:24 Doom mythmetadatalookup: mythmetadatalookup[19691]: I MetadataDownload metadatacommon.cpp:1105 (ParseMetadataItem) Result Found, Season 0 Episode 0
Mar 17 00:15:24 Doom mythmetadatalookup: mythmetadatalookup[19691]: I MetadataDownload metadatadownload.cpp:186 (run) Returning Metadata Results: The Daily Show with Trevor Noah 0 0
Mar 17 00:15:27 Doom mythmetadatalookup: mythmetadatalookup[19691]: N CoreContext main.cpp:167 (main) MythMetadataLookup run complete.
```

It seems odd that for a Manual Record rule a timestamp is always passed in for the -N invocation. I've searched through the code a bit, but haven't figured out why that happens. I tested all three television data grabbing scripts: tmdb3tv.py, ttvdb.py, and tvmaze.py and none of them seem to understand what to do with a timestamp argument.
I have run _tvmaze.py -N inetref episode-title_ and _tvmaze.py -D inetref year episode-num_ options on the command line, and they appear to be working properly.

SteveErl:
A copy of tvmaze.py was accidentally dropped into the top level mythtv directory. I've removed that bogus version. The real one is still around in mythtv/programs/scripts/metadata/Television/tvmaze.py

SteveErl:
The checks which are failing involve the TestProgramInfo test. I've run this test in my branch, and it works fine. But on a clone of 'master', it fails.

```
********* Start testing of TestProgramInfo *********
Config: Using QtTest library 5.9.5, Qt 5.9.5 (x86_64-little_endian-lp64 shared (dynamic) release build; by GCC 7.4.0)
PASS   : TestProgramInfo::initTestCase()
PASS   : TestProgramInfo::programFromVideo_test()
PASS   : TestProgramInfo::movieComparison_test()
XFAIL  : TestProgramInfo::programToStringList_test() Intentionally changed title.
   Loc: [moc/../test_programinfo.h(491)]
...
PASS   : TestProgramInfo::SubstituteMatches2_test(dirfile)
FAIL!  : TestProgramInfo::test_toMap() 'actualValue == expected[key]' returned FALSE. (Key 'starttime': actual '7:00 PM' is not expected '8:00 PM')
   Loc: [moc/../test_programinfo.h(618)]
FAIL!  : TestProgramInfo::test_toMap() 'actualValue == expected[key]' returned FALSE. (Key 'starttime': actual '7:00 PM' is not expected '8:00 PM')
   Loc: [moc/../test_programinfo.h(618)]
PASS   : TestProgramInfo::cleanupTestCase()
Totals: 23 passed, 2 failed, 0 skipped, 0 blacklisted, 22ms
```
The failure always involves a time difference of 1 hour, so I suspect it might involve Daylight Savings Time in some way. I wonder if something was checked into the 'master' branch, after my branch was created, which is causing this breakage.

SteveErl:
It looks like the changes made to
`mythtv/libs/libmyth/test/test_programinfo/test_programinfo.h`
_Add testing of programinfo::ToMap results_
on 3/13/2021 may be responsible for the failure. The Daylight Savings spring forward was on 3/14/2021.

linuxdude42:
Can you rebase to current master and try again. I added a test to TestProgramInfo on Tuesday that used localtime instead of UTC to check its results. That's likely the source of your problem.I committed fixed code this morning that performs the tests in UTC.

SteveErl:
@linuxdude42 - I merged in your fix, and it worked. Thanks.

SteveErl:
I've just added support for the -N <title> <date time> syntax, which mythmetadatalookup frequently tries to run, but which has always failed until this implementation. Note that ttvdb.py and tmdb3tv.py do not yet support this syntax. Here's an example:

```
/usr/share/mythtv/metadata/Television/tvmaze.py -l en -a US -N "CBS Sunday Morning" "2021-03-21 08:00:00"
<metadata>
  <item>
    <title>CBS News Sunday Morning</title>
    <subtitle>Episode 12</subtitle>
    <description></description>
    <season>2021</season>
    <episode>12</episode>
    <inetref>15779</inetref>
    <collectionref>15779</collectionref>
    <language>en</language>
    <releasedate>2021-03-21</releasedate>
    <popularity>86.0</popularity>
    <year>2021</year>
    <runtime>90</runtime>
    <studios>
      <studio name="CBS"/>
    </studios>
    <people>
      <person name="Jane Pauley" character="Host" job="Actor"/>
      <person name="Charles Kuralt" character="Host" job="Actor"/>
      <person name="Charles Osgood" character="Host" job="Actor"/>
      <person name="Lee Cowan" character="Substitute Host" job="Actor"/>
      <person name="Erin Moriarty" character="Erin" job="Actor"/>
    </people>
    <images>
      <image type="coverart" url="https://static.tvmaze.com/uploads/images/original_untouched/237/592585.jpg" thumb="https://static.tvmaze.com/uploads/images/medium_portrait/237/592585.jpg"/>
      <image type="coverart" url="https://static.tvmaze.com/uploads/images/original_untouched/53/132964.jpg" thumb="https://static.tvmaze.com/uploads/images/medium_portrait/53/132964.jpg"/>
      <image type="coverart" url="https://static.tvmaze.com/uploads/images/original_untouched/90/226497.jpg" thumb="https://static.tvmaze.com/uploads/images/medium_portrait/90/226497.jpg"/>
      <image type="fanart" url="https://static.tvmaze.com/uploads/images/original_untouched/221/553290.jpg"/>
    </images>
  </item>
</metadata>
```

SteveErl:
_The text format parser tossed out part of my previous comment. The first sentence should have been:_ 
I've just added support for the -N **_title_** **_datetime_** syntax.

rcrdnalor (changes requested):
MythTV's Python Bindings have a 'datetime' module which is time-zone aware.
Since tvmaze.py uses already these bindings, you should replace the
occurrences of 'datetime' and 'dateutil' with the mythtv module 'datetime'.
Thus, you do not need conditional imports anymore.

SteveErl:
I'm trying to replace the normal datetime and dateutil libraries with MythTV's dt.py, but it's difficult to complete the operation.
```
import datetime
datetime.datetime.strptime()
```
becomes
```
from MythTV import datetime
datetime.strptime()
```
and
```
import dateutil
dateutil.tz.gettz()
```
becomes
`datetime.localTZ()`
but I haven't yet found a way to convert
`dateutil.tz.gettz(show_tz)`
This service transforms a string like "America/New_York" into a timezone instance which can be used in a replace() function call. Do you know if there is something equivalent in the MythTV 'datetime' module?

rcrdnalor:
To get a timezone instance from "America/New_York", you can use the class "posixtzinfo" from MythTV bindings,
but this class is well hidden:
`
$ python3
>>> from MythTV import datetime
>>> from MythTV.utility.dt import posixtzinfo
>>> now = datetime.now()
>>> other_tz = posixtzinfo("America/New_York")
>>> other_time = now.astimezone(other_tz)
>>> now.strftime("%Y-%m-%d %H:%M:%S")
'2021-03-27 13:45:33'
>>> other_time.strftime("%Y-%m-%d %H:%M:%S")
'2021-03-27 08:45:33'
>>> 
`

SteveErl:
Thanks for that example. I tried converting to using the MythTV datetime, and it almost worked, but the following call was always failing.
`episodes = tvmaze.get_show_episodes_by_date(inetref, dtInTgtZone)`
That routine works when called with a real datetime argument, but fails when passed a MythTV datetime argument. Is there a way to convert from a MythTV datetime to a normal datetime? Does the get_show_episodes_by_date() need to be changed to allow a MythTV datetime argument?
While looking at the tvmaze_api.py file, I found that it depends on the real datetime and dateutil libraries, so my addition of those libraries into tvmaze.py didn't pull in a **new** library requirement. I think I'll check in my latest updates, not using MythTV datetime, and then look further into the switch.

I've got a number of enhancements to handle situations where there are multiple matches on subtitle, and to better search for a subtitle match when there are multiple shows with a matching title. For example
`tvmaze.py -l en -a US -N "Bulletproof" "Episode 4"`
There are matches in two different seasons.
```
tvmaze.py -l en -a US -N "The Flash" "Double Vision"
tvmaze.py -l en -a US -N "The Flash" "To Catch a Blue"
```
In the original version, the "best" title match was the 2014 version of "The Flash", which didn't have any episodes matching the given subtitles. In the above examples, subtitle matches are in the 1990 show, and 1967 show. I've added in code to properly find the match.

rcrdnalor:
Yes, you are right, your changes did not introduce a 'new' dependency to python-dateutil, it was the grabber itself. I've overlooked that, see #347. 
Could you please update and extend the doc-tests at `tvmaze_tests.txt` with tests according your latest code change?
The script tvmaze.py has a switch named `--doctest`.
I will look into the api `get_show_episodes_by_date()` then and report back.

SteveErl:
@rcrdnalor Thanks for removing the dependence on dateutil in tvmaze_api.py, and thanks for your usage example. I've merged your update into my branch, and have removed use of dateutil from my version of tvmaze.py. I do still have `from datetime import timedelta`. I looked through the other python scripts and found many of them using the exact same syntax, so I think it's okay.

I found some weaknesses in the string matching algorithm, so I modified it. The Levenshtein distance is the minimum number of single character edits required to change one string into another. It's valuable information in
trying to find a best match. However, it's problematic to make decisions solely upon this distance. If the strings are really long or really short, a simple rule, such as "accept any distance < 6" can provide poor results. If we're comparing "Risk" and "Call", the Levenshtein distance is 4, but it's a terrible match. If we're comparing "In Sickness and in Health" and "Sickness and Health", the Levenshtein distance value is 6. This would fail the given rule, even though the strings are really a pretty good match. 60% of the characters are in common. To establish how good of a match we have, we'll use the Levenshtein distance to figure out the ratio of matching characters in the longer string divided by the length of the longer string. I call this 'match_quality'. This ratio ranges from 0 (no match at all) to 1.0 (a perfect match). This performs well for short, average, and long strings.

I also built in some flexibility when the subtitle is really a date-time value. If the start time of the recording falls anywhere within the duration of the initial broadcast, it will match. This helps in the case that a user starts their recording a little late. I also allow a match when the recording starts a little early. When I was testing, I found I had a Manual Record rule which started at 8:00pm, but the network actually started the show at 8:01.

I had to update tvmaze_tests.txt because the results to the very first -M test changed since I last checked in the expected results a few days ago. I found the forced verbose mode to be annoying because it was generating thousands of lines of output. So, I modified the --doctest handling to only be verbose when the --debug option is also provided on the command line. This way, the user can see just the test which failed, without being flooded by the output of all the tests which passed.

jpoet (commented):
Sorry I have not analyzed this change. Does it fix the situation where a subtitle (e.g. Episode 1) shows up in more than one season? If so, does it fix it even when you are recording an earlier season after later seasons have aired?

If the season # can not be _uniquely_ identified, it should not be modified.

SteveErl:
@jpoet I've updated the tvmaze.py grabber to return a list of all the episodes with a matching subtitle. MythTV currently picks the one at the top of the list, which will be the one in the most recent season. If you want to do something else, you could code up an alternative, but there's really no way to know which one is the "correct" one, without specifying the season. For example:
`tvmaze.py -N "The Moodys" "Episode 1"`
will return :
```
<metadata>
  <item>
    <title>The Moodys</title>
    <subtitle>Episode 1</subtitle>
    <description>The MOODYS picks up with SEAN SR. contemplating retirement - road trips across the country in an RV with ANN, while SEAN JR. takes over the family HVAC business. In reality, however, Ann has no intention of leaving her new psychology practice, and Sean Jr. continues to brew up schemes to get rich quick. BRIDGET, who is in the midst of a divorce, returns home while her fancy new townhouse is being renovated; and DAN leaves Brooklyn for Chicago to live with his girlfriend, CORA, only to realize the relationship is moving too quickly. Once again, the Moodys are all under the same roof.</description>
    <season>2</season>
    <episode>1</episode>
    ...
```
...
```
  <item>
    <title>The Moodys</title>
    <subtitle>Episode 1</subtitle>
    <description>Meet THE MOODYS, a tight-knit, but slightly dysfunctional family of five, all of whom gather in their hometown of Chicago for the "perfect" holiday. The family includes SEAN SR.; his wife, ANN; and their three grown children – DAN, the youngest of the siblings and the "creative one"; BRIDGET, the middle sister and the "overachiever"; and SEAN JR., the oldest sibling and "screw-up," still living at home with his parents. Freshly single, Dan returns to his childhood home for Christmas to find it's in various stages of disarray, with one functioning bathroom. As the extended family gathers for a welcome meal, Dan meets CORA, with whom he instantly clicks. There's only one problem: she's already dating his cousin.</description>
    <season>1</season>
    <episode>1</episode>
    ...
```

rcrdnalor:
@SteveErl In the meantime, this required check for a good pullrequest
is not valid anymore:
"commits are logically organised and have [good commit messages]"

Since this is your first pull request for mythtv, I took the freedom
to organize and summarize the various commits into the following
(see attached zip file):

0001-TVmaze-Adding-routine-to-collect-all-artwork.patch
0002-Get-more-artwork-and-favor-artwork-from-official-sou.patch
patches/0003-Adding-output-of-popularity-and-userrating.patch
patches/0004-Improved-search-for-subtitle.patch
patches/0005-Added-tvmaze.py-N-title-date-time-support.patch

If you agree with those patches, I will add your name and push 
them to master.
Otherwise, please close this pullrequest and open a new one.
(That is the usual way for good pull request reviews.)


@jpoet The grabber should report *all* data for a dedicated
[patches_#336.zip](https://github.com/MythTV/mythtv/files/6253155/patches_.336.zip)
[patches_#336.zip](https://github.com/MythTV/mythtv/files/6253157/patches_.336.zip)

 search request.
It is the duty of `libmythmetadata` to *filter* the available data
to it's needs. Extending/Fixing the functionality of `libmythmetadata`
is out of scope for this pull-request.

Happy Easter !

SteveErl:
@rcrdnalor Thanks for your suggested improvements, and for the reorganization. I hadn't added an example of **-N title datetime** to tvmaze_test.txt because I was worried about the time localization factor. I see that you implemented a workaround by converting from a UTC time to the local time before invoking the test case. Very clever.

The patches_#336.zip file looks good to me. Please add my name and push it to master.

SteveErl:
After applying the final patch to update tmaze_tests.txt, I ran a test of `tmaze.py --doctest` and it reported a failure. I've been seeing the first test scenario failing frequently. This time, one expected item, "The Night Watchman" had disappeared and one unexpected item, "Dana White: Lookin' for a Fight" had appeared. The test `tmaze.py -M "night a"` has a vague search criteria with many (currently 10) partial matches. The response is really large, often changes due to modifications in the TvMaze database. Maybe it should be replaced with something more specific, so there isn't as much churn in the expected results?

```
tvmaze.py -M "night a"|grep title
    <title>A Clear Midsummer Night</title>
    <title>A Night of Exploration</title>
    <title>Night Stalker: The Hunt for a Serial Killer</title>
    <title>A Night at the Theatre</title>
    <title>A Night at The Classic</title>
    <title>A Night with My Ex</title>
    <title>Tales for a Halloween Night</title>
    <title>Boys on Film - A Night with Duran Duran</title>
    <title>Light as a Feather</title>
    <title>Dana White: Lookin' for a Fight</title>

tvmaze.py -M "Castle Rock" | grep title
    <title>Castle Rock</title>
    <title>Rick Stein's Taste of the Sea</title>

tvmaze.py -M "Nancy Drew" | grep title
    <title>Nancy Drew</title>
    <title>Nancy Drew</title>
    <title>The Hardy Boys/Nancy Drew Mysteries</title>

tvmaze.py -M "Orange is the" | grep title
    <title>Orange Is the New Black</title>
    <title>Orange is the New Brown</title>
```

SteveErl:
@rcrdnalor Github says that "This branch has conflicts that must be resolved", so I clicked the **Resolve conflicts** button. The only "conflicts" are a couple of blank lines in tvmaze.py and the test you added into tvmaze_tests.txt. I'm not sure why those couldn't be automatically resolved. Anyway, I went through the annoying steps of manually deleting lines until the conflicts were resolved, but Github provides no option to complete that resolution operation. Is this really an operation which has to be done through git on the command line? Do I have to resolve these conflicts before your patches can be merged into master?

mspieth:
Ive been thinking about this for a while since I wrote the tests for ttvdb. There should be a way to mock the http get with captured data from a real run. This way the tests will always be consistent. Not sure how to do this with doctest format tests though.

Monkey patch the requests get call to feed the data and then set the data pointer to the capture data perhaps. Also perhaps validate the request too in a similar way.

PR #357, opened 2021-04-26
Update mythpainter.cpp

ReallyJustAnotherBear:
add header entry for 
#include <QPainterPath>


jhoyt4:
@ctreleaven - you may want to look at this one.  It adds a CONFIG_INCLUDEPATH for macports and removes a quicktime framework path.  I'm not sure if it will help or hurt the v29 port in the macports repository.

jhoyt4:
@ReallyJustAnotherBear - are these updates macOS specific?  If so, you may want to update the subject line to be a bit more descriptive / less nebulous.

stuarta:
As this is specific for v29 and we are no longer maintaining this version, I'm going to close this as WONTFIX.

Issue #406, opened 2021-11-01
configure does not set optimization flag in CXXFLAGS


ulmus-scott:
* **Platform**: Xubuntu 21.04
`Linux htpc 5.11.0-38-generic #42-Ubuntu SMP Fri Sep 24 14:03:54 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux`

* **MythTV version**: master, fixes/31

* **Package version**: mythbuntu ppa for fixes/31

* **Component**: configure

### What steps will reproduce the bug?
in mythtv/ run:
```
./configure
make
```

or look at config.mak instead of running make

for ppa: look at DWARF debugging info, note no `-O` flag
`readelf -wi <binary> | grep DW_AT_producer`
### What is the expected behaviour? 
Optimization flag is set for c++ compilation.
Without any options, `-O2`, maybe `-O3`, definitely at least `-O1`.

NB: `CFLAGS` includes `-O3`
### What do you see instead?
No `-O` flag is set, so gcc uses its default of `-O0`, no optimizations are performed.

### Additional information
with `./configure --extra-cxxflags=-O1` note `-O1` now in `CXXFLAGS`

linuxdude42:
You can add "--toolchain=hardened" to the configure command to get " -O2 -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2" added to the c++ flags.

kmdewaal:
I think that this bug report really refers to how the packages are built, as it refers to mythbuntu ppa for fixes/31. I do agree that packages should be built optimized and if that is not done, it should be fixed.

However, this proposal changes the compilation behavior for all interactive builds (./configure; make) and not (only) for package builds.
My expected behavior for interactive builds is that the default --compile-type=profile has the default optimization -O0. It can be an omission that this is not set in CXXFLAGS but then it should be set to -O0 if it is to be set and not to -O2.

garybuhrmaster:
> I think that this bug report really refers to how the packages are built, as it refers to mythbuntu ppa for fixes/31. I do agree that packages should be built optimized and if that is not done, it should be fixed.

Every distro has their own set of standards for CFLAGS/CXXFLAGS for building (some set/export the variables, some have config/build macros, some use the -specs option to gcc, whatever).  It is up to the packagers for that distro to make sure that those standards are implemented, and if there are mistakes in the distros packaging, there should be tickets opened with that distro against the package.

ulmus-scott:
@kmdewaal 
[...]

PR #411, opened 2021-11-07
configure: enable optimizations, use system libudfread if available

ulmus-scott:
A crude first pass at enabling optimizations (`-O2`) for C++ files.  See: https://github.com/MythTV/mythtv/issues/406

Also, some minor cleanup in configure.

If pkg-config finds libudfread, use it.  It is probably better to remove it and libbluray and in configure git clone and use their actual configure scripts.

Version 1.1.1 is because V 1.1.0 had pkg-config as udfread.  Also, https://pkgs.org/search/?q=udfread doesn't list many, <s>which might mean it is packaged as part of libbluray (my guess).</s> Some systems lack packages for libudfread.

@linuxdude42 https://github.com/MythTV/mythtv/commit/60ae2ac4647e68b059820fb77d5be0cb724f483d#diff-fb1929a0befb6d7e932cde30a73b5f0bcbb897b046b9a0d7b833c9523efd6c31R6993 should this be `$gxx_major_version -lt 11`?  Using `-isystem` to add the include paths for Qt, etc., is probably a better solution, however.

##### Checklist


garybuhrmaster:
> If pkg-config finds libudfread, use it. It is probably better to remove it and libbluray and in configure git clone and use their actual configure scripts.

One cannot "git clone" during builds.  One cannot presume network connectivity for building software (many distributions explicitly disable network connections during builds to insure reproducibility).

ulmus-scott:
> One cannot "git clone" during builds. One cannot presume network connectivity for building software (many distributions explicitly disable network connections during builds to insure reproducibility).

OK.  I was only floating that idea since both libudfread and libbluray are unmodified copies in the repository and `git clone` is used in https://github.com/MythTV/mythtv/blob/master/platform/win32/w64-mingw32/MythBuild_MXE.sh (found while `git grep`ping).

linuxdude42:
On Sun, 2021-11-07 at 11:35 -0800, ulmus-scott wrote:
> > One cannot "git clone" during builds. One cannot presume network
> > connectivity for building software (many distributions explicitly
> > disable network connections during builds to insure
> > reproducibility).
> OK. I was only floating that idea since both libudfread and libbluray
> are unmodified copies in the repository and git clone is used in
> https://github.com/MythTV/mythtv/blob/master/platform/win32/w64-mingw32/MythBuild_MXE.sh
> (found while git grepping).

That's initial support for a mingw32/64 port. At some point I hope that
ends up like the android port, where you only have compile the
libraries once, and then can compile MythTV as often as you want.

David

garybuhrmaster:
> That's initial support for a mingw32/64 port. At some point I hope that ends up like the android port, where you only have compile the libraries once, and then can compile MythTV as often as you want. David

Feels like that WIP should eventually end up in the packaging repo (like the android build port). and, in addition, ideally, all the steps to build a builder would be in ansible (reproducible builders and builds are the best).

PR #441, opened 2021-12-30
configure: Rename lib…_external to system_lib…

ulmus-scott:
This also renames using_lib…_external to using_system_lib… and
CONFIG_LIB…_EXTERNAL to CONFIG_SYSTEM_LIB…

Added define HAVE_LIBBLURAY to make the code independent from any such further changes,
and removed the now unused mythconfig.h includes.

This naming is clearer since the code is in the external/ directory.  This prevents confusion between what these settings/flags actually do and the code being under external/.

### DIFF
--- a/mythtv/configure
+++ b/mythtv/configure
@@ -2001,9 +2001,9 @@ MYTHTV_CONFIG_LIST='
     v4l2prime
     valgrind
     x11
-    libexiv2_external
-    libbluray_external
-    libudfread_external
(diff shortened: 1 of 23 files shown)


linuxdude42:
I like the clarification of the new naming it, but you need to remove the reordering of the header includes and push a new version.

ulmus-scott:
@linuxdude42 I removed the reordering.

PR #430, opened 2021-12-13
cleanup configure

ulmus-scott:
This should be a better way of setting the C++ optimization flag.

##### TODO
- [ ] Decide on a satisfactory resolution to https://github.com/MythTV/mythtv/issues/406

PR #442, opened 2021-12-30
replace CONFIG_DARWIN with Q_OS_DARWIN; update deprecated Q_OS_MAC and Q_OS_MACX

ulmus-scott:
Was part of https://github.com/MythTV/mythtv/pull/430

Added a few new commits regarding `linux` vs `__linux__`

### DIFF
--- a/mythplugins/mytharchive/mytharchivehelper/main.cpp
+++ b/mythplugins/mytharchive/mytharchivehelper/main.cpp
@@ -30,11 +30,11 @@
 #include <sys/wait.h>  // for WIFEXITED and WEXITSTATUS
 #include <unistd.h>
 
-#include <mythconfig.h>
-#if CONFIG_DARWIN or defined(__FreeBSD__)
+#include <QtGlobal>
+#if defined(Q_OS_DARWIN) or defined(__FreeBSD__)
 #include <sys/param.h>
 #include <sys/mount.h>
-#elif __linux__
+#elif defined(__linux__)
 #include <sys/vfs.h>
 #endif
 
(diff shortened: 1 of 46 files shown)


ulmus-scott:
Some of the macOS/Darwin conditional code probably needs to be tested to see if it is still needed, e.g. https://github.com/MythTV/mythtv/blob/991d0f473226d68693b40bf2ae3a5b3c5b6cd461/mythtv/libs/libmyth/mythcontext.cpp#L286-L289

ulmus-scott:
Ignore https://github.com/MythTV/mythtv/pull/442/commits/4c56a8ba9b19e77f48376914335f837c1560afc9
Its not supposed to be in this pull request.

linuxdude42:
I like all the bits of this pull request, but it is a mashup of several unrelated things. Please refocus it specifically on the CONFIG_DARWIN-> Q_OS_DARWIN and similar directive changes like the linux->__linux__ and "if x" to "if defined(x)".

The change for reordering compat.h should be separate pull requests so it can have a proper discussion. I think it looks good, but I'd like to consider it separately from all the preprocessor directive changes.

In the future, changes like cleaning up doxygen, removing XRANDR, removing mythio.h, deleting unused code in mythmusic, and removing the math checks from configure should all be separate pull requests. I cherry-picked them this time, and am currently running them through my build system. If there aren't any problems I'll push them in an hour or so. 

The change to httprequest.cpp should be enhanced to remove all references to USE_SETSOCKOPT and submitted as a separate pull. That code was all commented out by 9846f8f6e3 in 2014 and can be removed.

Please don't gratuitously reorder the include file statements in the code.

ulmus-scott:
> I like all the bits of this pull request, but it is a mashup of several unrelated things. Please refocus it specifically on the CONFIG_DARWIN-> Q_OS_DARWIN and similar directive changes like the linux->**linux** and "if x" to "if defined(x)".
> 
> In the future, changes like cleaning up doxygen, removing XRANDR, removing mythio.h, deleting unused code in mythmusic, and removing the math checks from configure should all be separate pull requests. I cherry-picked them this time, and am currently running them through my build system. If there aren't any problems I'll push them in an hour or so.

My goal was to reduce use of mythconfig.h.  I was cleaning up the code as I saw it.  I agree the pull requests could have been more focused.

> The change to httprequest.cpp should be enhanced to remove all references to USE_SETSOCKOPT and submitted as a separate pull. That code was all commented out by [9846f8f](https://github.com/MythTV/mythtv/commit/9846f8f6e3eddcbf1e5d74c50d5f994bbe162a66) in 2014 and can be removed.

I'll look into it.

> The change for reordering compat.h should be separate pull requests so it can have a proper discussion. I think it looks good, but I'd like to consider it separately from all the preprocessor directive changes.

See https://github.com/MythTV/mythtv/pull/447

> Please don't gratuitously reorder the include file statements in the code.

Reordering the includes does serve a purpose; however, I agree they should probably be in a separate commit, maybe even a separate pull request.

linuxdude42:
I just pushed a version of this without the include file reordering.

PR #449, opened 2022-01-02
replace "#ifdef linux" with "#ifdef __linux__"

ulmus-scott:
Was part of https://github.com/MythTV/mythtv/pull/442


linuxdude42:
I pushed a fixed version of this patch.

PR #456, opened 2022-01-10
enable -Wundef for C++ code

ulmus-scott:
Disable -Wundef for external/libexiv2/ because of noise from `xmpsdk/include/*`.

See individual commits for details.


bennettpeter:
Merged into master

PR #457, opened 2022-01-11
GitHub Actions (macOS): persist ccache's cache between runs

ulmus-scott:
ccache does not appear to work on the macOS actions.  I will attempt to enable and verify it on the macOS workflows.

### DIFF
--- a/.github/workflows/buildmaster.yml
+++ b/.github/workflows/buildmaster.yml
@@ -28,13 +28,22 @@ jobs:
     - name: Setup build environment
       run: echo "MYTHTV_CONFIG=--prefix=${{ github.workspace }}/build/install --cc=${{ matrix.cc }} --cxx=${{ matrix.cxx }}" >> $GITHUB_ENV
       
-    - name: Check ccache
+    - name: Check ccache (linux)
       uses: actions/cache@v2
       with:
         path: ~/.ccache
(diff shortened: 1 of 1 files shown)


ulmus-scott:
on macOS:
(default) cache_dir = /Users/runner/Library/Caches/ccache

https://github.com/MythTV/mythtv/blob/master/.github/workflows/buildmaster.yml#L34
Is wrong for macOS.

Waiting for the [second Action](https://github.com/MythTV/mythtv/actions/runs/1680670476) to complete.

ctreleaven:
If a macOS user has installed ccache via MacPorts, your assumption will be invalid--the cache_dir will be ~/.ccache.  I'm not sure where Homebrew or Fink place the cache.  I think the correct thing to do is to ask ccache:

`ccache --get-config cache_dir`

Craig

ulmus-scott:
@ctreleaven This only effects the GitHub Actions builds, not anyone working on a checked out copy on their local machine.

The GitHub builders use Homebrew:
https://github.com/MythTV/mythtv/blob/630afc10487ffdd7c9c9f7572cf7a78f39cb1d8b/.github/workflows/buildmaster.yml#L52

This was the easiest way I saw to enable ccache on the macOS GitHub Actions builders.

ulmus-scott:
I decided to force the use of `CCACHE_DIR=$HOME/.ccache` instead of relying on fragile packaging defaults.

PR #460, opened 2022-01-11
Use C++ <random> instead of Qt or POSIX

ulmus-scott:
This also removes the conditional compilation due to the Qt random number generator functions added in 5.10.

##### Checklist

### DIFF
--- a/mythplugins/mythmusic/mythmusic/bumpscope.cpp
+++ b/mythplugins/mythmusic/mythmusic/bumpscope.cpp
@@ -3,6 +3,7 @@
 
 #include <compat.h>
 #include <mythlogging.h>
+#include <mythrandom.h>
 
 // This was:
 // Bump Scope - Visualization Plugin for XMMS
@@ -409,7 +410,7 @@ bool BumpScope::draw(QPainter *p, const QColor &back)
         if (m_ixo > ((int)m_width / 2) || m_ixo < -((int)m_width / 2))
         {
             m_ixo = (m_ixo > 0) ? (m_width / 2) : -(m_width / 2);
-            if (random() & 1)
+            if (rand_bool())
             {
                 m_ixd = (m_ixd > 0) ? -1 : 1;
                 m_iyd = 0;
@@ -425,7 +426,7 @@ bool BumpScope::draw(QPainter *p, const QColor &back)
         if (m_iyo > ((int)m_height / 2) || m_iyo < -((int)m_height / 2))
         {
             m_iyo = (m_iyo > 0) ? (m_height / 2) : -(m_height / 2);
-            if (random() & 1)
+            if (rand_bool())
             {
                 m_ixd = (m_ixd > 0) ? -1 : 1;
                 m_iyd = 0;
@@ -445,14 +446,14 @@ bool BumpScope::draw(QPainter *p, const QColor &back)
             rgb_to_hsv(m_color, &m_ih, &m_is, &m_iv);
             m_wasColor = 1;
 
-            if (random() & 1)
+            if (rand_bool())
             {
-                m_ihd = (random() & 1) * 2 - 1;
+                m_ihd = rand_bool() ? -1 : 1;
                 m_isd = 0;
             }
             else
             {
-                m_isd = 0.01 * ((random() & 1) * 2 - 1);
+                m_isd = rand_bool() ? -0.01 : 0.01;
                 m_ihd = 0;
             }
         }
@@ -468,16 +469,16 @@ bool BumpScope::draw(QPainter *p, const QColor &back)
                 m_ih = 0;
             if (m_ih < 0)
                 m_ih = 359;
-            if ((random() % 150) == 0)
+            if (rand_bool(150))
             {
-                if (random() & 1)
+                if (rand_bool())
                 {
-                    m_ihd = (random() & 1) * 2 - 1;
+                    m_ihd = rand_bool() ? -1 : 1;
                     m_isd = 0;
                 }
                 else
                 {
-                    m_isd = 0.01 * ((random() & 1) * 2 - 1);
+                    m_isd = rand_bool() ? -0.01 : 0.01;
                     m_ihd = 0;
                 }
             }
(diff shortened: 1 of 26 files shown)


kmdewaal:
This pull request does not only change the random function but it has also changes in the range of the random numbers in some places.

garybuhrmaster:
> This also removes the conditional compilation due to the Qt random number generator functions added in 5.10

Another alternative, which would appear more consistent with the MythTV coding standards documentation, would be that after the (imminent) release of v32 to advocate for bumping the minimum QT level(*) and cleaning up the ifdef'ery if the core devs are too annoyed with such.

But that would be something for master-next (v33-Pre) after checking what the various target OS's now have as their shipping version (I suspect 5.12 could be the new minimum, but I have not checked all the various target platforms).

(*) As I recall, with the current master, the reasoning for Qt 5.9 as a minimum was that at the time a few of the targeted OS's were at 5.9.  I think all of those OS's have either aged out of the targeted platform list, or are otherwise no longer supported for master (due to other requirements).

ulmus-scott:
@linuxdude42 accepted the change in range when he cherry-picked https://github.com/MythTV/mythtv/commit/760b38e706d441d04b329ada4d37ca61f61e2f47

@kmdewaal 
https://github.com/MythTV/mythtv/blob/8041f3f86989f9b2dc182f216c8296790e54e420/mythtv/libs/libmythtv/recorders/rtp/packetbuffer.cpp changing from 56 to 64 bits should have no effect.  However, I'm not sure PacketBuffer::FreePacket is correct because the lower 32 bits could overflow changing the upper 32.

Non-critical random wait times:
https://github.com/MythTV/mythtv/pull/460/files#diff-af72ee43895be54744dea803e57183f8ce7024a273902223b0af7009d23bc182
https://github.com/MythTV/mythtv/pull/460/files#diff-e934210f17d1d9c72e2200d9a833903d437da6408f366e73674fe564a5225514
https://github.com/MythTV/mythtv/pull/460/files#diff-1a52eb678c06743f5296042f36c82e40ee1402936f5fad255b3254aaf3b08b00
https://github.com/MythTV/mythtv/pull/460/files#diff-35e3f8dde951028c54b25561fe927967204c4148c9f88a95f276619ea8f95514

https://github.com/MythTV/mythtv/pull/460/files#diff-11bdaa792aface98719cb1be95d15cd68e31fce4489e517869a9121e13793eed More variance in exact wait value.

Other changes:
https://github.com/MythTV/mythtv/pull/460/files#diff-d9c5f12dbe3a9cc5f9a06a2af578836a58d8764036b8bf365bc1fa8d2b210076 This was wrong; it should have been `% 256`.  Incidentally, `% 255` produces a non-uniform distribution.

@garybuhrmaster I added most of the ifdefs in https://github.com/MythTV/mythtv/commit/760b38e706d441d04b329ada4d37ca61f61e2f47

stuarta:
> https://pkgs.org/search/?q=Qtbase It appears CentOS 7 and Ubuntu 18.04 LTS are still Qt 5.9.
> 
> Ubuntu 20.04 LTS is 5.12. Ubuntu 22.04 LTS should be 5.15 like everything else.
> 
> Of note, [Qt 5.12](https://www.qt.io/blog/qt-5.12.12-released) is EOL as of December.

Centos 7 has already been dropped from master.
We won't explicitly support Ubuntu-18.04 LTS once 22.04 LTS is out.

ulmus-scott:
The POSIX random() function has now been replaced with MythRandom.

ulmus-scott:
PacketBuffer::FreePacket no longer has any (reasonable) risk of errors from overflowing.

linuxdude42:
These changes look fine to me. I keep going back to look at the [packetbuffer.cpp](https://github.com/MythTV/mythtv/pull/460/files#diff-78c75c50677a92c64f87a31aa12c6440c86981feaf43475c7a8093f0802ebe1c) change because it does produce a different result than the original, but if the comment in packetbuffer.h is correct then this appears to be a better solution. The old solution filled the bottom 56 bits of ```m_next_empty_packet_key``` with a random number, where an increment (potentially the very first one) could overflow from the bottom 32 bits into the top 32 bits. This would cause ```PacketBuffer::FreePacket``` to stop caching packets with the old top 32 bits and only cache packets with the new top 32 bits. (Not that this would cause a problem, it would just churn packets for a while.) The new solution delays this churn basically indefinitely.

It does seems to me though that ```PacketBuffer::FreePacket``` never actually frees anything, and thus is a memory leak. Shouldn't packets that aren't added to ```m_empty_packets``` be deleted?

ulmus-scott:
> These changes look fine to me. I keep going back to look at the [packetbuffer.cpp](https://github.com/MythTV/mythtv/pull/460/files#diff-78c75c50677a92c64f87a31aa12c6440c86981feaf43475c7a8093f0802ebe1c) change because it does produce a different result than the original, but if the comment in packetbuffer.h is correct then this appears to be a better solution. The old solution filled the bottom 56 bits of `m_next_empty_packet_key` with a random number, where an increment (potentially the very first one) could overflow from the bottom 32 bits into the top 32 bits. 

I think the expected value of the initial packet buffer size would be 2^31 on average before overflowing and increasing in size to 2^32 after overflowing. 

> This would cause `PacketBuffer::FreePacket` to stop caching packets with the old top 32 bits and only cache packets with the new top 32 bits. (Not that this would cause a problem, it would just churn packets for a while.) The new solution delays this churn basically indefinitely.

On average it never happened before either (2^31).

> 
> It does seems to me though that `PacketBuffer::FreePacket` never actually frees anything, and thus is a memory leak. Shouldn't packets that aren't added to `m_empty_packets` be deleted?

If you look at [udppacket.h](https://github.com/MythTV/mythtv/blob/d47928088bc0cd3ab5b16fa183a865b0911da584/mythtv/libs/libmythtv/recorders/rtp/udppacket.h) and https://github.com/MythTV/mythtv/blob/d47928088bc0cd3ab5b16fa183a865b0911da584/mythtv/libs/libmythtv/recorders/rtp/packetbuffer.h#L35-L41

UDPPacket is a reference counted QByteArray, so the memory would be freed when all references have fallen out of scope.  It is not quite a memory leak because the packets can still be used in the future instead of creating new ones.  If the UDPPacket were a fixed size, I think it would make more sense.  A fixed size ring buffer might be a better choice instead of this.

Although, once the PacketBuffer goes out of scope, all of the memory should be freed.

So it is a change, but I don't think it's a significant one.  We could also leave that for later and just use 
```
MythRandom64() & ((UINT64_C(1) << 56) - 1)
```

.

ulmus-scott:
@linuxdude42 Some additional probabilities for overflowing:
The probability the old code overflowed given 2^N unprocessed packets was `P = 2^(-32 + N)`; e.g. N = 10 (1024 packets), P = 1 / 4 194 304 .

PR #443, opened 2021-12-30
Minor cleanup to audio code

ulmus-scott:
Was part of https://github.com/MythTV/mythtv/pull/430

The QtEndian conversion is not complete, but this was sufficient to remove libmythbase/bswap.h.

`git grep -nE BIGENDIAN -- :^*/external/**` still shows a few files.

PR #470, opened 2022-02-01
replace configure's ARCH... with Q_PROCESSOR_...

ulmus-scott:
This redoes part of https://github.com/MythTV/mythtv/pull/443 , so I will rebase that on top of this.

In mythdeinterlacer.cpp: x86_64 assumes <emmintrin.h> exists; however, I'm not sure if it would also be acceptable
to assume <arm_neon.h> exists when compiling for ARM, i.e. replace `HAVE_INTRINSICS_NEON` with `defined(Q_PROCESSOR_ARM)`.

It needs further investigation, but AudioOutputUtil::has_hardware_fpu()/has_optimized_SIMD() may no longer be necessary.  It is from https://github.com/MythTV/mythtv/commit/c0fc09267318eb8400e7d809c6ab698c5c325507 which reverts https://github.com/MythTV/mythtv/commit/2d67f9f5a384fab8aced0e3ed9f82065864f758d in 2010 (and references https://code.mythtv.org/trac/ticket/6569 ).

FFmpeg appears to have implemented [dialnorm](https://en.wikipedia.org/wiki/Dialnorm) in 2014 for [AC3](https://github.com/FFmpeg/FFmpeg/commit/12df9b9a151026c4b382df8852fad38165b49f95) and in 2016 for [EAC3](https://github.com/FFmpeg/FFmpeg/commit/96cd6f672e5d8c5d49b06de4f24376f36880fea8).

### DIFF
--- a/mythplugins/settings.pro
+++ b/mythplugins/settings.pro
@@ -41,11 +41,6 @@ QMAKE_CXXFLAGS += $$CXXFLAGS $$ECXXFLAGS
 profile:!win32:!macx:CONFIG += debug
 
 QMAKE_CXXFLAGS_RELEASE = $$OPTFLAGS -fomit-frame-pointer
-release:contains( TARGET_ARCH_POWERPC, yes ) {
-    QMAKE_CXXFLAGS_RELEASE = $$OPTFLAGS
-    # Auto-inlining causes some Qt moc methods to go missing
-    macx:QMAKE_CXXFLAGS_RELEASE += -fno-inline-functions
-}
 QMAKE_CXXFLAGS_RELEASE += $$PROFILEFLAGS
 
 QMAKE_CFLAGS += $$ARCHFLAGS
--- a/mythtv/libs/libmyth/audio/audioconvert.cpp
+++ b/mythtv/libs/libmyth/audio/audioconvert.cpp
@@ -35,11 +35,14 @@ extern "C" {
 #include "libswresample/swresample.h"
 }
 
+#include <QtGlobal>
+
+
 #define LOC QString("AudioConvert: ")
 
 #define ISALIGN(x) (((unsigned long)(x) & 0xf) == 0)
 
-#if ARCH_X86
+#ifdef Q_PROCESSOR_X86
 static int has_sse2 = -1;
 
 // Check cpuid for SSE2 support on x86 / x86_64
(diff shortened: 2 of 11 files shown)


ulmus-scott:
> 6\. Disabling a #if in [zoom_filter_xmmx.cpp](https://github.com/MythTV/mythtv/pull/470/files#diff-3b18e9f8da3eb16676491bd752d2691f31cf722773a9a6c6e1b1a3fc3285292b) ??

See the [commit message](https://github.com/MythTV/mythtv/pull/470/commits/0fdd49be31553e4296d56bf9e93cc2113d4dacb4): configure *always* defines ARCH_X86_64 (to either 0 or 1).

> 
> Referencing #443, that appears to repeat most of items 1,2,3,4 and also has:
> 
>     7. A different std::clamp
> 
>     8. Changing the way HAVE_INTRINSICS_NEON is et.
> 
>     9. Replacing uchar with uint8_t, and renaming clip_uchar to clip_uint8.

GitHub doesn't really have any way to do dependent pull requests.  https://github.com/MythTV/mythtv/pull/443 is the last six commits on top of this PR.

linuxdude42:
> > 6. Disabling a #if in [zoom_filter_xmmx.cpp](https://github.com/MythTV/mythtv/pull/470/files#diff-3b18e9f8da3eb16676491bd752d2691f31cf722773a9a6c6e1b1a3fc3285292b) ??
> 
> See the [commit message](https://github.com/MythTV/mythtv/pull/470/commits/0fdd49be31553e4296d56bf9e93cc2113d4dacb4): configure _always_ defines ARCH_X86_64 (to either 0 or 1).

This seems like a mistake caused by the difference between the two styles of config.h files. This file appears to be expecting the "#define FOO" or "// #undef FOO" style, vs. the "#define FOO 0/1" style that MythTV uses.

ulmus-scott:
This has been replaced by:

- [x] https://github.com/MythTV/mythtv/pull/487
- [x] https://github.com/MythTV/mythtv/pull/488
- [x] https://github.com/MythTV/mythtv/pull/489
- [ ] https://github.com/MythTV/mythtv/pull/491
- [ ] https://github.com/MythTV/mythtv/pull/492

ulmus-scott:
All part have now been merged separately.

PR #488, opened 2022-02-17
goom: re-enable zoom_filter_xmmx.cpp for x86_32

ulmus-scott:
and remove use of ARCH_...

MMX is a CONFIG_DEFINES in configure
https://github.com/MythTV/mythtv/blob/8cbe83862ccf2c467758e87c7d9d100176daf7c6/mythtv/configure#L7899-L7904

mythconfig.h always has ARCH_... defined.

Modified from commit in https://github.com/MythTV/mythtv/pull/470

### DIFF
--- a/mythtv/libs/libmythtv/visualisations/goom/zoom_filter_xmmx.cpp
+++ b/mythtv/libs/libmythtv/visualisations/goom/zoom_filter_xmmx.cpp
@@ -1,8 +1,6 @@
-#include "mythconfig.h"
-
 #include "goom/zoom_filters.h"
 
-#if defined(MMX) && !defined(ARCH_X86_64)
+#if defined(MMX) && !(defined(__x86_64) || defined(__x86_64__) || defined(__amd64) || defined(_M_X64))
 /* a definir pour avoir exactement le meme resultat que la fonction C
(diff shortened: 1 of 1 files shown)


linuxdude42:
This fails to compile on my debian 12 i386 builder.  This is on a VM running a Pentium III cpu.

$ uname -a
Linux dt-i386.dhampton.net 5.15.0-3-686-pae #1 SMP Debian 5.15.15-2 (2022-01-30) i686 GNU/Linux
cat /etc/os-release 
PRETTY_NAME="Debian GNU/Linux bookworm/sid"
NAME="Debian GNU/Linux"
ID=debian

The end of the compile log:

linking mythfrontend
/usr/bin/ld: ../../libs/libmythtv/libmythtv-33.so: undefined reference to `zoom_filter_xmmx(int, int, unsigned int const*, unsigned int const*, int const*, int const*, int, std::array<std::array<int, 16u>, 16u>&)'
collect2: error: ld returned 1 exit status
make[2]: *** [Makefile:497: mythfrontend] Error 1

linuxdude42:
I think I fixed the underlying problem. I'll commit this now.

PR #504, opened 2022-02-25
configure: echo ffopts

ulmus-scott:
for debugging failing launchpad armhf build

The compile check succeeds in mythtv/configure but fails in FFmpeg/configure.  This should help identify the cause and is good info to print regardless.


ulmus-scott:
Only use the first commit.

linuxdude42:
I applied the first commit.

Issue #505, opened 2022-02-26
Fail to build for armhf


marillat:
Debian armhf unstable

commit https://github.com/MythTV/mythtv/commit/75cf5040eff2aa6b7dc075462311de7c277b9375 is not enough to fix the arm CPU detection.
The main configure use the same broken detection. Applying the same patch fix this issue.

````
--- a/configure
+++ b/configure
@@ -4308,9 +4308,10 @@ elif enabled arm; then
         fi
     }
 
-    [ "$cpu" = generic ] && cpu=$(probe_arm_arch)
-
     case $cpu in
+        generic)
+            subarch=$(probe_arm_arch | sed 's/[^a-z0-9]//g')
+        ;;
         armv*)
             cpuflags="-march=$cpu"
             subarch=$(echo $cpu | sed 's/[^a-z0-9]//g')
````

PR #507, opened 2022-02-26
configure: fix --cpu=host for Intel Celeron and Pentium

ulmus-scott:
class processors.

This is necessary because they don't support AVX, but e.g. gcc's
-march=haswell enables AVX instructions; however, the Haswell
Pentium G3258 does not support AVX, which -march=native correctly
selects.

cpu cannot be set to native, since then the conditional checks,
for e.g. fast cmov, are not executed.

This should fix https://github.com/MythTV/mythtv/issues/505 in master and can be cherry-picked cleanly into fixes/31 and fixes/31.

### DIFF
--- a/mythtv/configure
+++ b/mythtv/configure
@@ -3405,7 +3405,7 @@ if test "$cpu" = host; then
     esac
 
     if test "${cpu:-host}" != host ; then
-        cpuflags_native="-march=native"
+        cpuflags_native="-march=native -mtune=native"
     else
         enable proc_opt_old
         cpu="$arch"
--- a/mythtv/configure
+++ b/mythtv/configure
@@ -6382,6 +6382,7 @@ ffmpeg_optenableif() {
 }
 
 if [ -n "$cpuflags_native" ]; then
+    cpu_detected="$cpu"
     cpu="host"
 fi
 
@@ -6485,7 +6486,14 @@ echo "install prefix            $prefix"
 echo "runtime prefix            $runprefix"
 echo
 if enabled cpu_override && test x"$cpu" != x"generic"; then
-    echo "CPU override              $arch $subarch ($cpu)"
+    if [ -n "$cpuflags_native" ]; then
+        echo "CPU override              $arch $subarch (cpu=host (native) is $cpu_detected)"
+        if test x"$processor" != x"" ; then
+            echo "                          [$processor]"
+        fi
+    else
+        echo "CPU override              $arch $subarch ($cpu)"
+    fi
 elif test x"$processor" != x"" ; then
     echo "CPU                       $arch $subarch ($processor)"
 else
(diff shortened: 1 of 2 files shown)


pcrow:
Executive summary: It looks good for me.

Details:

I saved the patch to /etc/portage/patches/media-tv/mythtv and emerged mythtv-31.0_p20210731.  [This is a Gentoo-ism for applying patches during a build while continuing to use the package management system.]  I'm compiling on a Haswell Celeron.  The configuration set HAVE_LRINT correctly (the first symptom of the problem encountered).  Running 'ps' during the build, I see -march=native on the gcc command lines, but on the cc1 command lines, I see -march=haswell, but then with tons of explicit flags including -mno-avx, so I think this is just exposing some irrelevant gcc internals, and it's doing the right thing.  I didn't do any further code testing, but it appears to me that it solves the problem.  (I also didn't closely examine the code changes.)

Doing a `grep -r haswell` in the build tree while running, I do notice:
```
mythtv/libs/libmythbase/moc/moc_predefs.h:#define __tune_haswell__ 1
mythtv/libs/libmythbase/moc/moc_predefs.h:#define __haswell__ 1
mythtv/libs/libmythui/moc/moc_predefs.h:#define __tune_haswell__ 1
mythtv/libs/libmythui/moc/moc_predefs.h:#define __haswell__ 1
mythtv/libs/libmythservicecontracts/moc/moc_predefs.h:#define __tune_haswell__ 1
mythtv/libs/libmythservicecontracts/moc/moc_predefs.h:#define __haswell__ 1
```
(More show up as the build progresses.)  I don't see anywhere where those defines are used unless I'm missing something, so I assume it's harmless.

ulmus-scott:
> Running 'ps' during the build, I see -march=native on the gcc command lines, but on the cc1 command lines, I see -march=haswell, but then with tons of explicit flags including -mno-avx, so I think this is just exposing some irrelevant gcc internals, and it's doing the right thing. 

I suspect this has to do with how `-march=native` probably works: the `gcc` program converts it to options for `cc1` which itself knows nothing about `-march=native`.

> Doing a `grep -r haswell` in the build tree while running, I do notice:
> 
> ```
> mythtv/libs/libmythbase/moc/moc_predefs.h:#define __tune_haswell__ 1
> mythtv/libs/libmythbase/moc/moc_predefs.h:#define __haswell__ 1
> mythtv/libs/libmythui/moc/moc_predefs.h:#define __tune_haswell__ 1
> mythtv/libs/libmythui/moc/moc_predefs.h:#define __haswell__ 1
> mythtv/libs/libmythservicecontracts/moc/moc_predefs.h:#define __tune_haswell__ 1
> mythtv/libs/libmythservicecontracts/moc/moc_predefs.h:#define __haswell__ 1
> ```
> 
> (More show up as the build progresses.) I don't see anywhere where those defines are used unless I'm missing something, so I assume it's harmless.

moc is the Qt Meta-Object Compiler, so that is probably fine.

Issue #511, opened 2022-03-03
Regression: video pauses at end of file instead of exiting


ulmus-scott:
* **Platform**: Xubuntu 21.10 `Linux htpc 5.13.0-30-generic #33-Ubuntu SMP Fri Feb 4 17:03:31 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
`

* **MythTV version**: v33/master

* **Package version**: n/a

* **Component**: mythfrontend

### What steps will reproduce the bug?

Play a recording or video and let it play until the end of file.

### How often does it reproduce? Is there a required condition?

Always.

### What is the expected behaviour?

Playback exits.

### What do you see instead?

Playback is paused at end of file.  `E  decoding error End of file (-541478725)` is shown in the log.  Detecting EOF may be a solution.

Playback exits with `Esc` as expected or after pausing and playing the video (hitting `space` twice).

### Additional information

git bisect yielded https://github.com/MythTV/mythtv/commit/754d740d00455aaacbe7926520816dc72830b67e @gigem

gigem:
I can't reproduce this with either current master or current fixes/32.  Both exit at the end of playback as expected.

ulmus-scott:
In the frontend settings: Video -> Playback -> General Playback -> "Action on playback exit" is "exit without saving" and "prompt at end of recording" is unchecked.

On master I can trigger this every time.  However, when making changes to the code to aid harmonizing mpegts-mythtv, in some instances it exited as expected, so this may be a race condition.  (I am testing on a 2 core Intel Pentium G3258.)

From v32:
[mythfrontend.20220303031531.66427.log](https://github.com/MythTV/mythtv/files/8174666/mythfrontend.20220303031531.66427.log)
Note: "End of file" occurs only once the first time, but the second time I played that file I paused and then played at the end and now it appears three times.

A race condition is highly likely: 
[mythfrontend.20220303032300.66593.log](https://github.com/MythTV/mythtv/files/8174730/mythfrontend.20220303032300.66593.log)
Exits the first time, but not the second for the same video file.  Fails to exit a different recording.

[Samples](https://drive.google.com/drive/folders/1vbvazwdu1K6Vb5elOGPid16djuYOMKM3) from https://code.mythtv.org/trac/ticket/13557 : file "a" was race-y, file "b" may also be race-y.

bennettpeter:
I tried this and I observe different behavior.
[...]

PR #512, opened 2022-03-03
mythplayer: fix exiting playback at EOF

ulmus-scott:
Fixes https://github.com/MythTV/mythtv/issues/511

This also applies cleanly to fixes/32.

##### Checklist

### DIFF
--- a/mythtv/libs/libmythtv/mythplayer.cpp
+++ b/mythtv/libs/libmythtv/mythplayer.cpp
@@ -732,14 +732,20 @@ bool MythPlayer::PrebufferEnoughFrames(int min_buffers)
     if (min_buffers)
         wait = m_videoOutput->ValidVideoFrames() < min_buffers;
     else if (GetEof() != kEofStateNone)
-        wait = false;
+        wait = false; // other code detects EOF to exit playback, so keep playing
     else if (abs(m_ffrewSkip) > 1)
         wait = !m_videoOutput->ValidVideoFrames();
     else
         wait = !m_videoOutput->EnoughDecodedFrames();
 
-    if (wait)
+    if (!wait)
     {
+        if (!m_avSync.GetAVSyncAudioPause())
+            m_audio.Pause(false);
+        SetBuffering(false);
+        return true;
+    }
+
         SetBuffering(true);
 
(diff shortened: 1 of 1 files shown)


gigem:
I won't pull this change to master as as as it conflicts with larger, in-progress changes I already have that will be commited soonish.  I will include the ultimate fix then or soon after.

I like the change with the short circuit return as the logic is clearer.  However, it is provably wrong for the function to return true when there are no frames available.  It is likely the clearing of m_buffering that is the important change here.  Please verify that doing that and still returning false in fixes/32 fixes the problem.  If it doesn't, please keep digging for the underlying bug.

ulmus-scott:
> However, it is provably wrong for the function to return true when there are no frames available.

If it works, and is documented in a comment, it seems fine to me.

PrebufferEnoughFrames() appears to only be called in one place: https://code.mythtv.org/doxygen/mythplayerui_8cpp_source.html#l00646

> It is likely the clearing of m_buffering that is the important change here. Please verify that doing that and still returning false in fixes/32 fixes the problem. 

It doesn't (tested on master):
With:
```
    else if (GetEof() != kEofStateNone)
    {
        // other code detects EOF to exit playback, so keep playing
        SetBuffering(false);
        return false;
    }
```

at EOF, playback is frame-by-frame with each play/pause.

With:
```
    else if (GetEof() != kEofStateNone)
    {
        // other code detects EOF to exit playback, so keep playing
        SetBuffering(false);
        return m_videoOutput->ValidVideoFrames() != 0;
    }
```

it stops playback (presumably on the last frame) as before my fix.

With:
```
    else if (GetEof() != kEofStateNone)
    {
        // other code detects EOF to exit playback, so keep playing
        if (!m_avSync.GetAVSyncAudioPause())
            m_audio.Pause(false);
        SetBuffering(false);
        return m_videoOutput->ValidVideoFrames() != 0;
    }
```

It sometimes works, but mostly not.


> If it doesn't, please keep digging for the underlying bug.

I'll have to try tracing the call(s) to PrebufferEnoughFrames() more fully to determine if/where additional EOF code should go.

gigem:
Fixed differently.

PR #517, opened 2022-03-09
de-duplicate signal handler code

ulmus-scott:
The only functional change is mythavtest and mythscreenwizard now use the logSigHup() signal handler, like the rest of the programs, instead of just ignoring it.

SIGHUP is hangup, which is raised when the terminal that launched it closes (or the session logs out, if I understand correctly), and by default terminates the application https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/signal.h.html .

By ignoring SIGHUP the application outlives the terminal that launched it, which makes more sense as a default when the program was launched as a background process with `&`.

logSigHup() calls FileLogger::reopen for each FileLogger, which presumably closes the logging to the terminal and fails to reopen it.
https://github.com/MythTV/mythtv/blob/1a9204f41e09c3a4652e1eec058d820c184a29f2/mythtv/libs/libmythbase/loggingserver.cpp#L150-L159

##### Checklist

### DIFF
--- a/mythtv/libs/libmythbase/signalhandling.cpp
+++ b/mythtv/libs/libmythbase/signalhandling.cpp
@@ -44,9 +44,7 @@ static void sig_str_init(void)
         sig_str_init(i, qPrintable(QString("Signal %1").arg(i)));
 }
 
-QList<int> SignalHandler::s_defaultHandlerList;
-
-SignalHandler::SignalHandler(QList<int> &signallist, QObject *parent) :
+SignalHandler::SignalHandler(QObject *parent) :
     QObject(parent)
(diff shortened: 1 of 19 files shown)


linuxdude42:
Based on the comment on line 151, this code exists so that you can use logrotate(8) on the logfiles generated by mythfrontend and mythbackend. All the other MythTV applications perform a single function and exit, so there would never be a need to rotate their log files.

ulmus-scott:
I don't see any harm in having all the programs use the same SIGHUP handler.  Do you?

linuxdude42:
None at all.

PR #526, opened 2022-03-19
MythPlayer::HasReachedEof(): short circuit the special cases

ulmus-scott:
This has no functional change, but is clearer logic.

### DIFF
--- a/mythtv/libs/libmythtv/mythplayer.cpp
+++ b/mythtv/libs/libmythtv/mythplayer.cpp
@@ -661,16 +661,13 @@ void MythPlayer::DiscardVideoFrames(bool KeyFrame, bool Flushed)
 
 bool MythPlayer::HasReachedEof(void) const
 {
-    EofState eof = GetEof();
-    if (eof != kEofStateNone && !m_allPaused)
-        return true;
     if (GetEditMode())
         return false;
     if (m_liveTV)
         return false;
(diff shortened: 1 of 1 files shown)


mythtv/libs/libmythtv/mythplayer.cpp
```diff
@@ -661,16 +661,13 @@ void MythPlayer::DiscardVideoFrames(bool KeyFrame, bool Flushed)
...
-    if (eof != kEofStateNone && !m_allPaused)
-        return true;
     if (GetEditMode())
         return false;
     if (m_liveTV)
         return false;
     if (!m_deleteMap.IsEmpty() && m_framesPlayed >= m_deleteMap.GetLastFrame())
         return true;
-    return false;
+    return (GetEof() != kEofStateNone && !m_allPaused);
```
linuxdude42, line 670: This can return different results than before. Try the combination of eof is kEofStateDelayed, m_allPaused is false, and GetEditMode() is true. Before the change this would return true, after the change it returns false.

ulmus-scott: Good point, I'll have to look into it more deeply, but my intuition is that you wouldn't want to exit because of EOF while editing the recording.

It may also make sense to add:
```
if (GetEof() == kEofStateImmediate)
    return true;
```

I'll investigate further.

PR #533, opened 2022-04-01
reduce duplicate filenames (main.cpp and commandlineparser.*)

ulmus-scott:
by by renaming to `(program|plugin|libplugin).cpp` and `(program)_commandlineparser.*`.

This is intended to reduce confusion in the Doxygen docs and also any errors while compiling.

##### Checklist


linuxdude42:
Looks good to me except for one thing. Having three files named mythutil.cpp, mythutils.cpp and mythutils.h is confusing. Given that mythutils.cpp has only a single function, and that the content of mythutils.h is relevant to both c++ files, I think the best approach would be to just collapse them down to mythutil.{cpp,h}. I know that means moving mythutils.h back to mythutil.h.

ulmus-scott:
> Looks good to me except for one thing. Having three files named mythutil.cpp, mythutils.cpp and mythutils.h is confusing. Given that mythutils.cpp has only a single function, and that the content of mythutils.h is relevant to both c++ files, I think the best approach would be to just collapse them down to mythutil.{cpp,h}. I know that means moving mythutils.h back to mythutil.h.

I have now merged mythutil.cpp into main.cpp and then renamed main.cpp to mythutil.cpp.

PR #540, opened 2022-04-09
Base watch list ordering on series most recently warched or recorded

Glidos:
(Also fixes the problem, when determining which episode to display, of air
date taking priority over series and episode)

Before this change, the watch list used a complex ordering that was
designed to help users keep up to date with many series simultaneosly,
ensuring that they didn't get behind with some. It's cleverly designed,
but seems very specific to the author's particular viewing habits. It
doesn't account well for shows that you really don't want to watch just
now, but you think you might one day. Nor does it account for a wish to
wait for a whole series before binging.

This commit simplifies the design, taking the view that mostly the user
will want to continue with series that they've recently viewed, or one to
which a new recording has recently been added. The new ordering gives the
user more control: if you wish to make a series appear higher in the watch
list then just watch a few seconds of an episode; if you wish to make a
series appear lower then don't watch any episodes from it and it will
get pushed out by the series you are watching.

### DIFF
--- a/mythtv/programs/mythfrontend/playbackbox.cpp
+++ b/mythtv/programs/mythfrontend/playbackbox.cpp
@@ -3140,7 +3140,8 @@ void PlaybackBox::ShowActionPopup(const ProgramInfo &pginfo)
 
         m_popupMenu->AddItem(tr("Recording Options"), nullptr, createRecordingMenu());
 
-        if (m_groupList->GetItemPos(m_groupList->GetItemCurrent()) == 0)
+        if (m_groupList->GetItemPos(m_groupList->GetItemCurrent()) == 0 ||
+            m_groupList->GetItemCurrent()->GetData().toString() == m_watchGroupLabel)
         {
             m_popupMenu->AddItem(tr("List Recorded Episodes"),
                                  &PlaybackBox::ShowRecordedEpisodes);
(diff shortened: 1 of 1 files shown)


mythtv/programs/mythfrontend/playbackbox.cpp
```diff
@@ -1981,205 +1995,27 @@ bool PlaybackBox::UpdateUILists(void)
...
-                maxEpisodes[recid] = query.value(2).toInt();
-                avgDelay[recid] = query.value(3).toInt();
-
-                QDateTime next_record =
-                    MythDate::as_utc(query.value(4).toDateTime());
-                QDateTime last_record =
-                    MythDate::as_utc(query.value(5).toDateTime());
+
                 QDateTime last_delete =
                     MythDate::as_utc(query.value(6).toDateTime());
```
ijc, line 2000: Since you aren't using the other values any more I think you could simplify the SQL now to just `SELECT recordid, last_delete FROM record;` and this would become `value(1)`.

Glidos: Oh! Good point. I will sort that out.

ijc:
I've applied this patch to my frontend and it produces more sensible looking watch lists to me. At the least it is now choosing the first season/episode by airdate rather than recording date. When I watched a few seconds of a thing it bumped up the list.

I think the proposed logic/simplification makes sense. If nothing else it seems like a much better base for future refinement to me (in so far as it's much easier to figure out what it actually does...).

Apart from the opportunity for simplifying the SQL I mentioned above the code looks fine to me (FWIW).

mythtv/programs/mythfrontend/playbackbox.cpp
Glidos: Okay. I've simplified the SQL, as you suggested.

Glidos:
I've just added one more commit to this pull request: it makes "List Recorded Episodes" available from the watch list.

PR #542, opened 2022-04-09
Potential memory corruption problems in audiooutputbase

acediac:

##### Checklist

- [ x] contribution does not duplicate one of our [existing pull requests](https://github.com/MythTV/mythtv/pulls)
- [ x] contribution is in a branch rebased against [master](https://github.com/MythTV/mythtv)
- [ x] code compiles successfully without errors

### DIFF
--- a/mythtv/libs/libmyth/audio/audiooutputbase.cpp
+++ b/mythtv/libs/libmyth/audio/audiooutputbase.cpp
@@ -91,13 +91,18 @@ AudioOutputBase::~AudioOutputBase()
                 "~AudioOutputBase called, but KillAudio has not been called!");
 
     // We got this from a subclass, delete it
-    delete m_outputSettings;
-    delete m_outputSettingsRaw;
-    if (m_outputSettings != m_outputSettingsDigital)
+    // These all seem to be the same pointer, avoid freeing multiple times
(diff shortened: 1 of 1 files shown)


acediac:
I have created a separate request for this file because I think it needs further review.

There are 4 member variables which are pointers to AudioOutputSettings:
m_outputSettings
m_outputSettingsRaw
m_outputSettingsDigital
m_outputSettingsDigitalRaw

In the original code on lines 94-100 each set of "raw and non-raw" pointers are freed in the destructor, but GetOutputSettingsCleaned assigns a new instance of AudioOutputSettings to the "raw" version of the pointer (174-175), which then gets set to equal to the "non raw" version of the pointer (200-201) at the end of GetOutputSettingsUsers (which calls GetOutputSettingsCleaned), which in turn was called by InitSettings and (redundantly?) sets the member pointers again (133-134) so potentially creating a situation where the same pointer is deleted more than once.

We also have a situation on lines 194-196 where we create a new instance, and then assign another new instance (through GetOutputSettingsCleaned) without deleting the first instance.

I think my proposed code will address these issues, but to be honest I'm not sure at a higher level what the code is trying to achieve so these are only bandaid fixes.

jhoyt4:
I've just tested this on my Intel and M1 macs both running Monterey (12.3.1) and everything works for me.

Issue #428, opened 2021-12-08
enable use of system FFmpeg


ulmus-scott:
In order to use the system supplied FFmpeg, we need to:

## 1. eliminate/upstream our modifications
- for eliminating some of the modifications, see #416 
- for logical commits rebased onto FFmpeg/release/4.4, see https://github.com/ulmus-scott/FFmpeg/commits/rebase/4.4m3

Investigate subtitle changes/additions:
- [ ] IVTV subtitles: https://github.com/ulmus-scott/FFmpeg/commit/a611475157919bec39c4405b4e4318524f59be4e https://github.com/ulmus-scott/FFmpeg/commit/2acbe0e1c10c77a46ea87069fe09b455014436ea
- [ ] DVB_VBI duplicates AV_CODEC_ID_DVB_TELETEXT? https://github.com/ulmus-scott/FFmpeg/commit/7bbd4d109a424bceba8be703d7c4c7240dc5dbb4
- [ ] ATSC subtitles: https://github.com/ulmus-scott/FFmpeg/commit/db044425081d738da6f7ec430b7bde27baae29fe

MPEG-TS changes
- [ ] investigate what the changes actually **are** versus the ffmpeg version: https://github.com/ulmus-scott/FFmpeg/commit/008d757f65dc1d123165e63d6e0c52d8bd6886cc

## 2.  Only use the following FFmpeg headers (eliminate use of internal FFmpeg headers)
Headers installed by the -dev packages on Ubuntu:
### libavcodec/:
ac3_parser.h
adts_parser.h
avcodec.h
avdct.h
avfft.h
bsf.h
codec_desc.h
codec.h
codec_id.h
codec_par.h
d3d11va.h
dirac.h
dv_profile.h
dxva2.h
jni.h
mediacodec.h
packet.h
qsv.h
vaapi.h
vdpau.h
version.h
videotoolbox.h
vorbis_parser.h
xvmc.h

### libavdevice/:
avdevice.h
version.h

### libavfilter/:
avfilter.h
buffersink.h
buffersrc.h
version.h

### libavformat/:
avformat.h
avio.h
version.h

### libavutil/:
adler32.h
aes_ctr.h
aes.h
attributes.h
audio_fifo.h
avassert.h
avconfig.h
avstring.h
avutil.h
base64.h
blowfish.h
bprint.h
bswap.h
buffer.h
camellia.h
cast5.h
channel_layout.h
common.h
cpu.h
crc.h
des.h
dict.h
display.h
[...]

PR #416, opened 2021-11-20
ffmpeg cleanup

ulmus-scott:
This now contains multiple pull requests with the following changes:

I decided to drop the MythTV changes to hide the use of FFmpeg's internal header libavformat/url.h since it only includes public headers and it is more involved to replace its use and I have not fully investigated it.

Otherwise, the MythTV changes I have kept are in:
- [ ] https://github.com/MythTV/mythtv/pull/568 stylistic change
- [ ] https://github.com/MythTV/mythtv/pull/565 lavc/utils-mythtv.c
- [ ] https://github.com/MythTV/mythtv/pull/576 Miscellaneous MythTV changes
- [ ] https://github.com/MythTV/mythtv/pull/557 ByteReader
- [ ] https://github.com/MythTV/mythtv/pull/577 BitReader 

I have rebased this onto master and made the following changes:

- renamed ff_codec_type_string to AVMediaTypeToString in mythtv/libs/libmythtv/decoders/avformatdecoder.cpp
- dropped changes concerning the use of FFmpeg's internal header libavformat/url.h since it only includes public headers and it is more involved to replace its use and I have not fully investigated it.
- only change to ffmpeg compared to the previous version is:
```
git diff ffmpeg-cleanup-backup -- mythtv/external/FFmpeg
diff --git a/mythtv/external/FFmpeg/libavcodec/libavcodec.v b/mythtv/external/FFmpeg/libavcodec/libavcodec.v
index 3f3dc96da0..f1d5e5bc77 100644
--- a/mythtv/external/FFmpeg/libavcodec/libavcodec.v
+++ b/mythtv/external/FFmpeg/libavcodec/libavcodec.v
@@ -1,11 +1,6 @@
 LIBAVCODEC_MAJOR {
     global:
         av*;
-        ff_ue_golomb_vlc_code;
-        ff_golomb_vlc_len;
-        ff_se_golomb_vlc_code;
-        ff_fft_init;
-        ff_fft_end;
     local:
         *;
 };
```

----
I am currently splitting this into multiple pull requests:

- [ ] https://github.com/MythTV/mythtv/pull/557
- [ ] bitreader (shared context with bytereader)
- [x] https://github.com/MythTV/mythtv/pull/558
- [x] https://github.com/MythTV/mythtv/pull/559
- [ ] https://github.com/MythTV/mythtv/pull/565
- [ ] https://github.com/MythTV/mythtv/pull/566
- [ ] ...
[...]

PR #557, opened 2022-05-08
create ByteReader to eliminate use of FFmpeg modification

ulmus-scott:
by copying avpriv_find_start_code() into MythTV as ByteReader::find_start_code().

Refs: https://github.com/MythTV/mythtv/issues/428

Note: I have rewritten this function.  The only functional change is
00 00 01 00 01 XX no longer incorrectly returns a start code at
offset 7 that overlaps the start code at offset 4 if the start_code
input is not modified between the two calls.

FFmpeg only has one use that keeps the history,
(ff_)mpeg1_find_frame_end().

So, I have split the truncated start code logic into its own function
as I have now done for FFmpeg, which only used this functionality once.

Determining if it is necessary would require further analysis or
adding logging to see if truncated start codes are detected.

This was originally part of https://github.com/MythTV/mythtv/pull/416

##### Checklist

### DIFF
--- a/mythtv/external/FFmpeg/libavcodec/avcodec.h
+++ b/mythtv/external/FFmpeg/libavcodec/avcodec.h
@@ -4184,7 +4184,6 @@ int avcodec_is_open(AVCodecContext *s);
 /* MythTV */
 const char *ff_codec_id_string(enum AVCodecID codec_id);
 const char *ff_codec_type_string(enum AVMediaType codec_type);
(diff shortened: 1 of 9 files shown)


ulmus-scott:
For reference:
https://github.com/MythTV/mythtv/blob/97dee9420ae47215fdc6ca71a76ed8c9f27cdab5/mythtv/external/FFmpeg/libavcodec/utils.c#L1011-L1042

ulmus-scott:
Will be merged as part of https://github.com/MythTV/mythtv/pull/416

PR #603, opened 2022-07-08
fix Ubuntu 22.04 GitHub Actions

ulmus-scott:
First, edit `mythplugins/configure` to enable building MythNetvision.  Then, fix the dependencies in the workflow `.yml` file.

##### Checklist

### DIFF
--- a/.github/workflows/buildmaster.yml
+++ b/.github/workflows/buildmaster.yml
@@ -78,10 +78,16 @@ jobs:
       run: make test
       if: runner.os == 'Linux'
 
-    - name: Install plugin dependencies (linux)
+    - name: Install plugin dependencies (ubuntu-20.04)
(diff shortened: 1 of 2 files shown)


ulmus-scott:
I would like comment on the Python oauth vs oauthlib because more may need to be changed to work on Ubuntu 22.04.  That is, it compiles the C++ MythNetvision, but the Python component may not work.

`git grep -nE oauth` seems to suggest that only vimeo uses it.

stuarta:
The python3-oauth package should be available to ubuntu-22.04.
Issue opened with github

https://github.com/actions/virtual-environments/issues/5895

stuarta:
Superceded by https://github.com/MythTV/mythtv/commit/fb9457e9aa4de597c2bcd17aec3bfbe6f6da1bbf

stuarta:
No they are not. Strictly speaking I could have completely left out that package. I believe mythnetvision needs updating to a different python oauth package. python3-oauthlib seems to be the favoured replacement. The other option is also unmaintained

PR #604, opened 2022-07-12
Logging cleanup and add options to force long console log format

ulmus-scott:
It was asked on IRC if there was an option to make the console log output include the line numbers, etc. in the last month or so, although I couldn't find it again.

The first part replaces the use of the `write` system call, which may truncate the output.

The second part adds the new compile-time and run-time options to force the long log format.  I am open to other name suggestions for the command-line flag and its corresponding variables.

Also, if there are any preferences for a different long log format, now is the time to change it.

As usual, see commits for details.

##### Checklist


ulmus-scott:
The format is currently `time LOG_LEVEL [pid{/tid}] thread_name __FILE__:__LINE__:__FUNCTION__  message`.  I would like to add a string version of `VB_...` after LOG_LEVEL.

Also, I think replacing `__FUNCTION__` with `__PRETTY_FUNCTION__` (GCC) / `__FUNCSIG__` (MSVC) might be beneficial.

paul-h:
It probably goes without saying but we need to be careful not to add too much extra processing of each log message because with the debug level enabled the logs can get very verbose and if you don't have a supercomputer then the logging just bog down the system making it harder to debug the problem you are trying to look at.

ulmus-scott:
As the changes currently stand, the FileLogger has an extra function call to generate the log message because it is not inline, but I don't consider that a noticeable increase because of the two QString function calls which malloc.

logConsole() has an extra if statement.

`__FUNCTION__` versus `__PRETTY_FUNCTION__`  / `__FUNCSIG__` is just a different `const char*` the compiler has already created.

Adding the LOG_LEVEL would, in my opinion, be useful for narrowing down what verbose arguments are needed when starting from `all` or `most`.  However, this would involve a large number of independent if statements (64?) because it is a bitset, not an enumeration.

ulmus-scott:
Rebased onto master without change to force rebuild.

ulmus-scott:
Rebased onto master and fixed a typo in a Doxygen comment.

twitham1:
I have found it challenging to discover where in the huge code base certain things happen.  Seems like this feature could make that much easier.  I might not want larger logs in production but it would be a nice option when trying to find a bug or feature to tweak.

ulmus-scott:
Rebased onto master.  I don't know what would be necessary to always force the long log format when using CMake.

PR #619, opened 2022-08-03
unziputils: move zlib include to proper file

ulmus-scott:
unzip2.(h|cpp) does not use any zlib symbols, only unziputil.cpp does.


ulmus-scott:
@linuxdude42 Ping.  You put the zlib include in the wrong file, `unzip2.(h|cpp)` do not need it.

PR #629, opened 2022-08-30
libavformat/mpegts-mythtv.c: fix segmentation fault from customization

ulmus-scott:
ts->req_sid could be set to a garbage value > 8192 on the first pass,
so keep the prior value, which is what the code did when it used a
pmt_filter variable that was part of the MpegTSContext.

---

This fixes the segmentation fault of `mythffprobe` with John Pilkington's BBC Proms sample from the dev mailing list.


bennettpeter:
I cherry-picked the commit into master.

PR #635, opened 2022-09-12
GitHub Actions: update cache Action to v3

ulmus-scott:
The cache has not been working for some time, hopefully this will fix it.


ulmus-scott:
This should fix the GitHub Actions cache not working, which causes the Actions to take over an hour.  For some reason `actions/cache@v2` does not work.

GitHub rotates the caches based on age and each run creates a new cache entry due to using the commit hash in the key.

Since ccache handles what GitHub caches, creating a new entry for each run should be unnecessary; however, I not sure how https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache would interact with this.

The other changes were made to test whether the cache worked, which it does.

PR #641, opened 2022-09-26
Revert change to use BitReader

ulmus-scott:
@jpoet I don't know how long it will take me to debug BitReader, so here is that change reverted.

##### Checklist

- [x] contribution does not duplicate one of our [existing pull requests](https://github.com/MythTV/mythtv/pulls)
- [x] contribution is in a branch rebased against [master](https://github.com/MythTV/mythtv)
- [x] code compiles successfully without errors
- [x] code follows the [MythTV Coding Standards](https://www.mythtv.org/wiki/Coding_Standards)
- [x] documentation added/updated/removed where necessary
- [x] commits are logically organised and have [good commit messages](https://chris.beams.io/posts/git-commit)

PR #642, opened 2022-09-26
Fix BitReader

ulmus-scott:
@jpoet The first three commits are fixes, then more unit tests, and some minor tidying up.

skip_bits() was broken for non aligned byte skips when the cache had data in it.  The new unit tests satisfy me that it works correctly now, but testing would be appreciated.

##### Checklist

### DIFF
--- a/mythtv/libs/libmythtv/bitreader.h
+++ b/mythtv/libs/libmythtv/bitreader.h
@@ -45,7 +45,7 @@ class BitReader
     }
     ~BitReader() = default;
 
-    void skip_bits(unsigned n = 1)
+    void skip_bits(unsigned n)
     {
         if (m_cacheSize > n)
         {
@@ -83,7 +83,7 @@ class BitReader
     }
     bool next_bit() { return get_bits(1) == 1; }
     /// @brief Read 0-32 bits.
-    uint32_t get_bits(unsigned n = 1)
+    uint32_t get_bits(unsigned n)
     {
         uint32_t ret = show_bits(n);
         skip_bits(n);
--- a/mythtv/libs/libmythtv/bitreader.h
+++ b/mythtv/libs/libmythtv/bitreader.h
@@ -254,15 +254,16 @@ inline void BitReader::refill_cache(unsigned min_bits)
     {
         int shift = 64 - m_cacheSize;
         int bits = CHAR_BIT - m_bitIndex;
-        m_cache |= int64_t(*m_buffer & ((1 << bits) - 1)) << (shift - bits);
         if (shift > bits)
         {
+            m_cache |= int64_t(*m_buffer & ((1 << bits) - 1)) << (shift - bits);
             m_bitIndex   = 0;
             m_buffer++;
             m_cacheSize += bits;
         }
         else
         {
+            m_cache |= int64_t(*m_buffer & ((1 << bits) - 1)) >> (bits - shift);
             m_bitIndex  += shift;
             m_cacheSize += shift;
         }
--- a/mythtv/libs/libmythtv/bitreader.h
+++ b/mythtv/libs/libmythtv/bitreader.h
@@ -187,14 +187,6 @@ class BitReader
 #endif
     }
 
-    static constexpr uint64_t mask_upper(unsigned bits)
-    {
-        // protect against undefined behavior
-        if (bits ==  0) return  UINT64_C(0);
-        if (bits >  64) return ~UINT64_C(0);
-        return ~((UINT64_C(1) << (64 - bits)) - 1);
-    }
-
     static constexpr uint64_t get_upper_bits(uint64_t val, unsigned bits)
     {
         // protect against undefined behavior
@@ -206,7 +198,7 @@ class BitReader
         {
             return val;
         }
-        return (val & mask_upper(bits)) >> (64 - bits);
+        return val >> (64 - bits);
     }
 
     void refill_cache(unsigned min_bits);
--- a/mythtv/libs/libmythtv/bitreader.h
+++ b/mythtv/libs/libmythtv/bitreader.h
@@ -47,14 +47,16 @@ class BitReader
 
     void skip_bits(unsigned n = 1)
     {
-        if (m_cacheSize >= n)
+        if (m_cacheSize > n)
         {
             m_cache <<= n;
             m_cacheSize -= n;
         }
         else
         {
+            n -= m_cacheSize;
             m_cacheSize = 0;
+            m_cache      = 0;
             m_bitIndex += n;
             int quotient = m_bitIndex / CHAR_BIT;
             m_bitIndex %= CHAR_BIT;
--- a/mythtv/libs/libmythtv/bitreader.h
+++ b/mythtv/libs/libmythtv/bitreader.h
@@ -72,7 +72,7 @@ class BitReader
         }
         return uint32_t(get_upper_bits(m_cache, n));
     }
-    uint32_t show_bits64(unsigned n)
+    uint64_t show_bits64(unsigned n)
     {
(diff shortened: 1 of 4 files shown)


warpme:
Scott,
I gave try applying above patches on current master.
Testing one 4k sat tv channel (FashionTV) I started to have totally errant durations. 

What i done:
I started livetv with this channel. 
I can't skip back in livetv. progress is exactly the same like duration.
I exited livetv and went to recordings.
Playback of recording from 1m20s livetv session has 4h50m duration on osd.
reverting patches makes channel working again*

(*) - this channel has issue of slow drift at livetv: looking at osd when livetv plays I see - at 60sec livetv duration shows 40sec progress. but livetv recording has duration ok

ulmus-scott:
Piotr,

How does the behavior compare with https://github.com/MythTV/mythtv/pull/641 reverting the change to BitReader?

warpme:
Scott,
I done some testing with backend running:

1. current master
2. current master + #642
3. current master + #641

all test went with 4k FasionTV sat channel.
tested aspects:

- -skip/forward in livetv
- -correct duration on OSD when livetv plays
- -correct progress/duration on OSD when playing recording from livetv session   

1. current master:

- -skip/forward in livetv: works
- -correct duration on OSD when livetv plays: drift about 10..20sec on progress at 60sec duration
- -correct progress/duration on OSD when playing recording from livetv session: OK

2/3. current master + #642; current master + #641

- -skip/forward in livetv: can't skip/forward (no effect on screen)
- -correct duration on OSD when livetv plays: progress is exact duration all the time
- -correct progress/duration on OSD when playing recording from livetv session: 1m20sec livetv session shows as 4h30m rec

ulmus-scott:
Klaas,

I think none of the parsers would work properly if they hit the end, regardless of reading all `1`s (the padding) or all `0`s.

Besides, Piotr reported the same behavior with both `get_bits.h`/`golomb.h` and the fixed BitReader, so I expect the problem to be in the parsers themselves not the bit reader implementation that is used.

BitReader is also cached, but that shouldn't make a difference.

kmdewaal:
Scott,

It could just be that there is more than one problem....

The bisecting done by John Poet shows that the HEVC parser starts to fail
after the introduction of the BitReader.
This can only be explained by differences between the BitReader and the
FFmpeg functions.
The stated goals of the BitReader, as I understand it, are to avoid using
internal FFmpeg functions and to remove compilation warnings.
These goals can be achieved also when the BitReader behaves exactly the
same as the FFmpeg functions that it replaces.
The HEVC parser should then give exactly the same results, independent of
whether the BitReader or the FFMpeg functions are used.

The reporting of Piotr shows that there are probably other issues as well
in the parsers, independent of which bit reader implementation is used.

Klaas.

ulmus-scott:
> The bisecting done by John Poet shows that the HEVC parser starts to fail after the introduction of the BitReader. 
>
> This can only be explained by differences between the BitReader and the FFmpeg functions. 

Yes, because BitReader::skip_bits() is broken, which is what this pull request fixes.  I am waiting for confirmation from John that it works since I have no AVC or HEVC sources.

> The stated goals of the BitReader, as I understand it, are to avoid using internal FFmpeg functions and to remove compilation warnings. 

Yes, to avoid using internal FFmpeg headers/functions.  You can disable the compilation warnings, which I proposed in https://github.com/MythTV/mythtv/pull/641 if we decide to revert the switch to BitReader.

> These goals can be achieved also when the BitReader behaves exactly the same as the FFmpeg functions that it replaces. 
>
> The HEVC parser should then give exactly the same results, independent of whether the BitReader or the FFMpeg functions are used. 
>
> The reporting of Piotr shows that there are probably other issues as well in the parsers, independent of which bit reader implementation is used.

I agree.

warpme:
Scott, Klass,

My expierience (limited) playing with ffmpeg + consumers (apps) of ffmpeg: if there is ffmpeg function realising desired functionality - go and use it at (almost) and expense instead of try alternative implementation of it.

I have at lest 2 major args:
1. any internal change within ffmpeg can be real pain to follow (ffmpeg devs are not publishing internals + they are not wiling to spent their time on anybody who wantst to reinvent wheel they invented)
2. they have way bigger view/experience about variety/quirks of/in content which ffmpeg needs to understand/support so for sure they can implement needed functionality in best way (i'm not underestimating yours skills. I'm simply trying to say:  ffmpeg devs are dealing with media streams/codecs/quirks/spec deviations daily and it is not easy to beat them on this field...)

warpme:
Guys
1 Q:
I was playing only with backend.
remote frontend was always current master.
hope this is not distorting concluding regarding issue....

ulmus-scott:
Piotr,

If the source was HEVC I would not expect the frontend to matter.  If it was AVC, it might matter since AVFormatDecoder uses AVCParser.

jpoet:
> Piotr,
> 
> If the source was HEVC I would not expect the frontend to matter. If it was AVC, it might matter since AVFormatDecoder uses AVCParser.

Even with HEVC, the frontend version does seem to matter. With PR#642 my HEVC 1080p59.94 recordings now show the correct duration, but there is still something broken on the frontend...

I can run either PR#642 or 31ee7f9e64 for the backend, but I get different results on the frontend OSD (Steppes themes) for the audio codec when the source is E-AC3:
With  31ee7f9e64 on the frontend I see "E-AC3"
With PR#642 (or any commit after 31ee7f9e64) on the frontend I see an empty space where the audio codec info should be.

If the source is just AC3, then that is displayed correctly, just not E-AC3.

Scott, I will upload a sample with E-AC3 encoded audio.

ulmus-scott:
With your E-AC3 sample, git bisect shows 8cb2003b8e22a098e23f213d39eb43e5c9df92fd is the culprit.  Looking at it again, the lower case change is significant and glaringly obvious.  It is not obvious, however, where that icon comes from, but I'll keep investigating which change actually broke it.  (That commit can't be reverted directly because of the subsequent commit removing the customization to FFmpeg that it was using.)

Please compare the behavior of this pull request to that of https://github.com/MythTV/mythtv/pull/641 due to other unrelated intervening changes.

jpoet:
With the appropriate change to the Steppes theme, my HEVC recordings are now behaving correctly with this PR.

warpme:
Guys,
Just fyi: I applied #642 on current master+frontend.
Issue reported in https://github.com/MythTV/mythtv/pull/642#issuecomment-1259616433 is still present.
I have also frequent segfaults in mythpreviewgen on backend when I'm trying to watch recordings from livetv.
For me current solution for #642 is more regression than fix.

ulmus-scott:
Piotr,

The current master is broken, so either this pull request or https://github.com/MythTV/mythtv/pull/641 need to be merged.

It seems like there are issues with the parsers independent of the bit reader implementation.  I need logs (mythbackend -v record:debug --logpath=SOME_PATH) and a stack trace for a crash to try and diagnose the issue.

ulmus-scott:
@warpme Without logs, a stack trace, and a comparison of that bad behavior to #641 I will recommend this be merged and for you to open a new issue with logs and a stack trace.

Are these issues only effecting the FashionTV channel?

warpme:
Scott,
Very brief testing shows that issue is only on FashionTV channel.
If you believe/know #642 fixes issue and FashionTV is another issue - sure. pls merge.
I will provide separate ticket for FashionTV.
br

ulmus-scott:
@jpoet New commits only tweak the unit tests.  You can merge this pull request.

Piotr's FashionTV issue appears limited to that one channel and is independent of the bitstream reader implementation.  The problem is therefore likely in the parsers.  We will fix that issue separately after this has merged and Piotr has created a new issue with more information (logs, etc.).

PR #643, opened 2022-09-26
Manual Recordings: restore date time subtitle

ulmus-scott:
Apparently removed in 5f6697ecfaf9bd7175aaf99b3ef07f884d5af65c

##### Checklist

### DIFF
--- a/mythtv/programs/mythbackend/scheduler.cpp
+++ b/mythtv/programs/mythbackend/scheduler.cpp
@@ -3789,7 +3789,9 @@ void Scheduler::UpdateManuals(uint recordid)
             query.bindValue(":STARTTIME", startdt);
             query.bindValue(":ENDTIME", startdt.addSecs(duration));
             query.bindValue(":TITLE", title);
-            query.bindValue(":SUBTITLE", subtitle);
+            query.bindValue(":SUBTITLE",
+                            !subtitle.isEmpty() ? subtitle :
+                            MythDate::toString(startdt, MythDate::kDatabase | MythDate::kOverrideLocal));
             query.bindValue(":DESCRIPTION", description);
             query.bindValue(":SEASON", season);
             query.bindValue(":EPISODE", episode);


ulmus-scott:
I rebased onto master and did some additional cleanup.

I think an additional RecordingType that allows recording on either weekends or weekdays may be useful and make daily actually be all days:
```
diff --git a/mythtv/programs/mythbackend/scheduler.cpp b/mythtv/programs/mythbackend/scheduler.cpp
index 69cf223c52..bb3b2c19c9 100644
--- a/mythtv/programs/mythbackend/scheduler.cpp
+++ b/mythtv/programs/mythbackend/scheduler.cpp
@@ -3767,7 +3767,14 @@ void Scheduler::UpdateManuals(uint recordid)
     case kDailyRecord:
         for (int i = 0; i < daysInWeek * weeksToSchedule; i++)
         {
-            if (startDate.dayOfWeek() < 6 && startDate.addDays(offset + i).dayOfWeek() >= 6)
+            startList.push_back(QDateTime(startDate.addDays(offset + i), startTime, Qt::LocalTime).toUTC());
+        }
+        break;
+    case kWeekdayWeekendRecord:
+        for (int i = 0; i < daysInWeek * weeksToSchedule; i++)
+        {
+            if ((startDate.dayOfWeek() <  6 && startDate.addDays(offset + i).dayOfWeek() >= 6) ||
+                (startDate.dayOfWeek() >= 6 && startDate.addDays(offset + i).dayOfWeek() <  6))
             {
                 continue;
             }

```
but I'm not sure what changes would be required elsewhere.

bennettpeter:
David Engel should review anything that affects scheduler

gigem:
I'm deferring to John on this one as he made the change to remove it.  As I understand it, the API allows the user to set a desired subtitle and description.  Mythfrontend does not.  A work around might be to add the date to program.subtitle only when neither record.subtitle nor record.description are set.

ulmus-scott:
@jpoet's change sets an empty description to the date; however, the ProgramInfo constructor mythfrontend uses sets the description equal to the title, so that is never seen if using only the frontend.

https://github.com/MythTV/mythtv/blob/3d3a2cc5fa7c0bb715ab41ab785a76341373b08d/mythtv/libs/libmythbase/programinfo.cpp#L827-L828

PR #646, opened 2022-09-30
Manual Recordings: add optional UI element to set subtitle

ulmus-scott:
If the element is absent, the behavior is unchanged.

I don't know how to test default and default-wide, but the changes are identical to the other themes.

##### Checklist


ulmus-scott:
For earlier versions of MythTV, the theme changes for other themes will result in a textedit widget that does not effect the resultant recording rule, which may be confusing.  However, this is still technically backwards compatible.

bennettpeter:
The same change should be made in the web app. The web app has all the capabilities of the frontend for scheduling.

ulmus-scott:
It is reasonable that the web app should have the same capabilities as the frontend for scheduling.  However, I have no idea where to start or where to look.

As an aside why are the .js files all one line?

bennettpeter:
Here are the basics of the web app:
https://www.mythtv.org/wiki/Web_Application_Development
https://www.mythtv.org/wiki/Web_Application
The js files in mythtv/html/apps/backend are compiled from the source in mythtv/html/backend/src subdirectories.
We don't edit or look at those compressed js files, that will drive you crazy.
Angular uses a programming language called typescript, which is javascript with type checking added. It gets compiled into compressed javascript in main.js. Following  Stuart's lead, we compile it locally and commit the compiled code so that the mythtv build  process does not have to compile it.

billmeek:
On Tuesday 04 June 2024 09:14:20 AM (-05:00), Peter Bennett wrote:

 > Here are the basics of the web app:
 > https://www.mythtv.org/wiki/Web_Application_Development
 > https://www.mythtv.org/wiki/Web_Application
 > The js files in mythtv/html/apps/backend are compiled from the source in 
mythtv/html/backend/src subdirectories.
 > We don't edit or look at those compressed js files, that will drive you 
crazy.
 > Angular uses a programming language called typescript, which is 
javascript with type checking added. It gets compiled into compressed 
javascript in main.js. Following  Stuart's lead, we compile it locally and 
commit the compiled code so that the mythtv build  process does not have to 
compile it.

I use git logx to view the MythTV commits. It's an addition to my 
.git/config file done like this:

   git config alias.logx "log -p ':(exclude)html/apps/backend/main.js'"

It only works from the mythtv directory itself. I haven't found an 
equivalent
for git grep (other than piping to grep -v '^html/apps/backend/main.js'
-- 
Bill

ulmus-scott:
Assuming all the files in `mythtv/html/apps/` are autogenerated, adding a `.gitattributes` file https://github.com/MythTV/mythtv/pull/646/commits/674244516445369d65ec66d7ef4ffa38150ef61a will prevent those files from cluttering git diff, grep (from any directory), or log (including old commits).  If some of the files are not autogenerated, the pattern could be changed or they could be added back with `text=auto diff`.

https://git-scm.com/docs/gitattributes#_using_macro_attributes
https://stackoverflow.com/questions/12707808/is-it-possible-to-ignore-certain-directories-by-default-when-using-git-grep

I don't know JavaScript or TypeScript, but I'll see if there is an obvious change.

ulmus-scott:
I think the changes should be in:
```
mythtv/html/backend/src/app/schedule/schedule.component.html:
mythtv/html/backend/src/app/schedule/schedule.component.ts
```
but I couldn't figure it out.

Regardless, the web app appears to still work.  This makes sense since it doesn't use the "Manual Record" ProgramInfo constructor.  "Manual Search" is similar, but not equivalent, to "Manual Record".

Some comments on the UI:
- Why does "Manual Search" show the description if it can't be edited?
- Why does Search Type "None" show boxes for the title, subtitle, and description that can't be interacted with?  It would make more sense if the popup didn't show anything below Search Type until a Search Type was selected.

bennettpeter:
> * Why does "Manual Search" show the description if it can't be edited?
> * Why does Search Type "None" show boxes for the title, subtitle, and description that can't be interacted with?  It would make more sense if the popup didn't show anything below Search Type until a Search Type was selected.

It uses the same html file for all types of schedule. The subtitle and description are read only because their values come from the program guide.

Once your changes are implemented I will make the same changes to the web app and leanfront.

Angular is complex and takes some time to learn.

ulmus-scott:
Should I open a new PR for the .gitattributes file https://github.com/MythTV/mythtv/commit/674244516445369d65ec66d7ef4ffa38150ef61a or will you cherry-pick it?

> Once your changes are implemented I will make the same changes to the web app and leanfront. Let me know when that happens or open a ticket.

I don't know about leanfront, but I don't think you have to wait for these changes to implement something similar in the web app.

bennettpeter:
> Should I open a new PR for the .gitattributes file [6742445](https://github.com/MythTV/mythtv/commit/674244516445369d65ec66d7ef4ffa38150ef61a) or will you cherry-pick it?
> 
> >

I have now committed .gitattributes so you can take that out of the pull request

bennettpeter:
> I don't know about leanfront, but I don't think you have to wait for these changes to implement something similar in the web app.

I will leave it to you to verify that it actually works and that the subtitle actually gets on to the recording. Once that is all done and working it will be easy to add it to the web app GUI and to leanfront.

ulmus-scott:
I rebased onto master, removing the now redundant commit to add `.gitattributes` files and moving the second commit editing the "default-wide" theme to before the theme version bump.

> I will leave it to you to verify that it actually works and that the subtitle actually gets on to the recording.

I don't understand what you mean.  When creating a manual recording in mythfrontend via Manage Recordings -> Schedule Recordings -> Manual Schedule, if the new subtitle text box is left empty, the behavior is unchanged and, if something is entered into the subtitle text box, the recording will now have the entered subtitle instead of an empty string.

bennettpeter:
> I don't understand what you mean. When creating a manual recording in mythfrontend via Manage Recordings -> Schedule Recordings -> Manual Schedule, if the new subtitle text box is left empty, the behavior is unchanged and, if something is entered into the subtitle text box, the recording will now have the entered subtitle instead of an empty string.

That is perfect.

ulmus-scott:
Rebased onto master to force rebuild.

Issue #650, opened 2022-10-12
FashionTV (4k HEVC Sat TV channel) fails to record after g84a9131


warpme:
* **Platform**: backend: Archlinux; frontend: MiniMyth2

* **MythTV version**: master g84a913147c

* **Package version**: self-build with myth-master 84a913147c

* **Component**: recording

### What steps will reproduce the bug?

record on FashionTV (4k HEVC Sat TV channel)

### How often does it reproduce? Is there a required condition?

always

### What is the expected behaviour?

correct recoding

### Additional information
This is known regression introduced by 84a913147c13092ca6b1e39df18f15d064cceb14  

examplery failed recording file: http://warped.inet2.org/fashiontv-failed1.ts http://warped.inet2.org/fashiontv-failed2.ts
exemplary good recording file (before 84a913147c): http://warped.inet2.org/fashiontv-good.ts

pls let me know if any other logs/captures are required to move forward with this regression

jpoet:
Hey Piotr, I guess I missed part of what you were saying about this....
Are you saying the resulting recording file is damaged by these changes? I thought it was just a problem with the seektable.
In theory, the resulting recording files should never vary as the bytes are just being written to the disk. The only thing that is being processed is the location of keyframes and metadata.

warpme:
John,

Yes.
To confirm - when I apply https://github.com/warpme/minimyth2/blob/master/script/myth-master/mythtv/files/0060-ticket460-revert-ulmus-scott-fix_BitReader.patch (patch just reverts essential change of #642 on backend) - recording from FassionTV return to be ok.

btw: "broken seektable" was my hypothesis after just brief testing i done with livetv at times i was initially testing on #642 and provide feedback (seektable theory was my guess - as on livetv i had constant progress 00:00 of 00:00 and I can't skip back/forward in such livetv session)

jpoet:
Just watching those videos using mpv, the "good" one is clean almost immediately. The "failed" one it takes mpv almost 60 seconds before it plays it clean. It behaves like "keyframes" are very few and far between, but they should not be a whole minute apart. I will have to look at the recorder to see if I can figure out why the recorded file is being affected.

Please provide backend logs with "-v channel,record" when a program on that channel is the only thing recording. I would like to see what it says as it tries to "lock on" to that channel.

jpoet:
Piotr is this a DVB source? It would be very helpful if you could go into
[...]

PR #657, opened 2022-11-01
HEVCParser: fix parsing of st_ref_pic_set( stRpsIdx )

ulmus-scott:
This caused the bitstream to become misaligned for the satelite channel FashionTV UHD, which then caused the framerate to become an incorrect value.

Fixes https://github.com/MythTV/mythtv/issues/650

---

I could only test with two HEVC samples, so further testing would be greatly appreciated.


kmdewaal:
Scott, well done! This does fix the FTV UHD channel and it does not cause regressions on other HEVC-coded channels.

Issue #654, opened 2022-10-24
Tvmaze metadata lookup failures


SteveErl:
* **Platform**: All

* **MythTV version**: fixes/32

* **Package version**:

* **Component**: metadatalookup

### What steps will reproduce the bug?

Using the TVmaze service for metadata lookup, there are rare cases where the 'runtime' field is empty, but 'averageRuntime' is not. This happened when looking up metadata for a TV show called "Everything's Trash". Note that the database entry for this show was subsequently updated to include the missing 'runtime' field. But on the night it was recorded, that field was empty.

For the same show, the early contents of the database entries had 'airdate' populated, but no 'airtime'. I'm guessing that at the time the data was populated, the broadcast time had not been decided yet. The missing 'airtime' caused the specific episode lookup to fail.

### How often does it reproduce? Is there a required condition?
It's rare, but when it happens, it causes metadata lookups to fail.

### What is the expected behaviour?

The lookup scripts should be more flexible to handle these sorts of issues. 
If the 'runtime' field is empty, try using the 'averageRuntime' field. If the 'averageRuntime' field is also empty, set a default runtime length. If the 'airtime' field is empty, look for an episode which matches the 'airdate' field.

### Additional information

Here's an example of the original data for "Everything's Trash":

{"id":60518,
"url":"https://www.tvmaze.com/shows/60518/everythings-trash",
"name":"Everything's Trash",
"type":"Scripted",
"language":"English",
"genres":[],
"status":"Running",
"runtime":null,
"averageRuntime":30,
"premiered":"2022-07-13",
"ended":null,
"officialSite":null,
"schedule": {"time":"","days":[]},
"rating": {"average":null},
"weight":94,

SteveErl:
I have a solution for this bug. Once I figure out how to push my solution to my GitHub branch, the fix will be available.

rcrdnalor:
Looking at your provided patch, this issue actually addresses four items:
 - Cleanup python imports of not needed modules
   No functional change.
 - Fall back on average runtime if the the episode's runtime is not provided
   I am not sure what improvement is behind this change. I prefer to not
   provide inaccurate/guessed data if no data are available, e.g.: the 
   hardcoded 60 mins for a duration which is not provided by tvmaze. 
 - Add robustness when fetching a non existent timestamp for an episode:
   Do not interrupt iteration of all found episodes matching a certain timestamp
   if one episode does not provide accurate data.
[...]

PR #662, opened 2022-11-11
Fix setting of TvMaze show runtime

SteveErl:
The script tvmaze/show.py had a line which retrieved the episode duration from the metadata provider, but then assigned it to **num_episodes**, which makes no sense. There is no code looking for **num_episodes**, but there is code looking for **runtime**, so the line has been fixed to assign to **runtime**.

Using the TVmaze service for metadata lookup, there are rare cases where the 'runtime' field is empty, but 'averageRuntime' is not. This happened when looking up metadata for a TV show called "Everything's Trash". Note that the database entry for this show was subsequently updated to include the missing 'runtime' field. But on the night it was recorded, and for several days later, that field was null.

Here's an excerpt of the original metadata for "Everything's Trash":

"id":60518,
"url":"https://www.tvmaze.com/shows/60518/everythings-trash",
"name":"Everything's Trash",
"type":"Scripted",
"language":"English",
"genres":[],
"status":"Running",
"runtime":**null**,
"averageRuntime":**30**,

To handle this type of scenario, when the 'runtime' field is null, read the duration from the 'averageRuntime' field. If both are null, then avoid trying to cast that value to an integer in tvmaze.py.

Refs: #654

##### Checklist

### DIFF
--- a/mythtv/bindings/python/tvmaze/show.py
+++ b/mythtv/bindings/python/tvmaze/show.py
@@ -47,7 +47,10 @@ class Show(object):
         self.language = show.get('language')
         self.genres = show.get('genres')
         self.status = show.get('status')
-        self.num_episodes = show.get('runtime')
+        self.runtime = show.get('runtime')
+        # Sometimes runtime is empty, but averageRuntime is not
+        if self.runtime is None:
+            self.runtime = show.get('averageRuntime')
         self.seasons = {}
         self._episode_list = []
         self.specials = {}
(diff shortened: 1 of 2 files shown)


ulmus-scott:
You should probably amend your commit message to include your original commit message:
```
Fix setting of TvMaze show runtime

When a TvMaze episode runtime field is missing, try using averageRuntime field.

If both are missing, avoid a crash by testing for a non-NULL episode duration.

Refs https://github.com/MythTV/mythtv/issues/654
```

PR #664, opened 2022-11-14
Default to last channel used in Manual Record rule

### LINKED ISSUE #655: Remember last channel used in Manual Record rule
**Is your feature request related to a problem? Please describe.**
When a user is entering a number of Manual Record rules in a row, it would nice if the previously used channel was remembered. For example, if you wanted to enter rules to record **Chicago Med**, **Chicago Fire**, and **Chicago P.D.**, which are all on the same channel, it would save time on the second and third rules, if the channel set from the first one was set as the initial channel value.

**Describe the solution you'd like**
When a Manual Record rule is entered, store the channel number such that it can be set as the default channel for the next Manual Record rule created.

**Describe alternatives you've considered**

SteveErl:
When a user is entering a number of Manual Record rules in a row, it would be nice if the previously used channel was remembered and used as the default for the current rule. For example, if you wanted to enter rules to record Chicago Med, Chicago Fire, and Chicago P.D., which are all on the same channel, it would save time on the second and third rules, to default to the common channel.

To implement this, when a Manual Record rule is entered, store the channel number such that it can be set as the default channel for the next Manual Record rule.

Fixes #655

##### Checklist

### DIFF
--- a/mythtv/programs/mythfrontend/manualschedule.cpp
+++ b/mythtv/programs/mythfrontend/manualschedule.cpp
@@ -56,6 +56,8 @@ bool ManualSchedule::Create(void)
 
     QString startchan = gCoreContext->GetSetting("DefaultTVChannel", "");
     QString chanorder = gCoreContext->GetSetting("ChannelOrdering", "channum");
+    QString lastManualRecordChan = gCoreContext->GetSetting("LastManualRecordChan", startchan);
+    int manStartChanType = gCoreContext->GetNumSetting("ManualRecordStartChanType", 1);
     ChannelInfoList channels = ChannelUtil::GetChannels(0, true, "channum,callsign");
     ChannelUtil::SortChannels(channels, chanorder);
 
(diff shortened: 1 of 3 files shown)


mythtv/programs/mythfrontend/manualschedule.cpp
```diff
@@ -75,6 +77,9 @@ bool ManualSchedule::Create(void)
         m_chanids.push_back(channels[i].m_chanId);
     }
 
+    if (last_manual_chan != -1)
+        m_channelList->SetItemCurrent(last_manual_chan);
+
```
ulmus-scott, line 82: https://www.mythtv.org/wiki/Coding_Standards#Spaces_and_braces " One-line `if` or `else` bodies should have braces."

mythtv/programs/mythfrontend/manualschedule.h
```diff
@@ -60,6 +60,7 @@ class ManualSchedule : public MythScreenType
     QString m_startString;
     QString m_chanidString;
 
+    static int last_manual_chan; /* Channel used in the last manual schedule */
```
ulmus-scott, line 63: https://www.mythtv.org/wiki/Coding_Standards#Class_and_variable_names This name should be `s_lastManualChan`, i.e. prefixed with `s_` and **lowerCamelCase**; however, I like the name `s_lastChannel` better since the manual is already implied by `ManualSchedule::`.  Or `s_lastChannelIndex` since it is an index.

(Feel free to ignore this one.) Personally, I find the separate declaration and definition syntax a little strange since you have to re-declare the type (I understand why, because the declared type could be incomplete).  Since we are using C++17, you can declare the static variable `inline` in the class with the definition.

ulmus-scott (commented):
This is a behavioral change and I agree it would be a slight quality of life improvement.

No locking on this global state should be fine since there should only ever be one ManualSchedule screen at one time.

ulmus-scott (approved):
I can't merge, but this looks good to me.

I'd have kept the comment on the declaration of `s_lastChannelIndex`, though, and probably would have made it a Doxygen comment using `///<`, or equivalently `/**< ... */`.

gigem:
Why are you needing to enter so many, manual recordings?  MythTV is designed to work best with good, guide data such as from EIT where it is available or from an external provider like Schedules Direct.  Manual recordings are there for the rare cases when the guide data is missing or wrong.

garybuhrmaster:
Based on past best practices, all behavioral changes in UI's (which may be a regression for others) MUST (IETF MUST) be based on a database setting, with the default left as it is.  I do not see such a DB setting.

SteveErl:
@garybuhrmaster : I've pushed a new commit which satisfies the suggested best practice.

PR #669, opened 2022-11-22
Set Subtitle in Change Recording Metadata Search

### LINKED ISSUE #667: Change Recording Metadata Search fails to set Subtitle
* **Platform**: All

* **MythTV version**: fixes/32, master

* **Package version**:

* **Component**: frontend

Create a Television recording without a Season, Episode, Subtitle, or Description. This should be a real program for which a TV metadata search will find a matching title. A **Manual Record** will generally accomplish this.

In the frontend, focus on this recording and then request an edit of the metadata. By default, this can be done by keying in the letter '**m**' to raise the Recording Options menu. From here, select '**Recording Options**' -> '**Change Recording Metadata**'.
MissScarletEdit

This will open the '**Edit Recording Metadata**' window. The **Title** field will be populated. The **Subtitle** field will be empty, and the **Season** and **Episode** values will be zero. There may be a generic **Description** for the TV show.

Now, specify the **Season** and **Episode** with valid values.
MissScarletPreSearch

Next click the **Search** button. This will run the metadata grabber to try to find all the specific metadata information for this particular episode. This does set a specific **Description**, but the **Subtitle** field never gets populated.

100% of the time.

The retrieved Subtitle should be written into the **Subtitle** field on the screen.
MissScarletFixedSearch

MissSarletFinal

### What do you see instead?
MissScarletPostSearch

MissScarletPostEdit

SteveErl:
After a metadata Search within the 'Edit Recording Metadata'
screen, the subtitle value is mistakenly written to the Description
field. Then that value is usually immediately overwritten by the
real description value.

This code change fixes the problem by properly
writing to the Subtitle field.

Resolves #667

##### Checklist

### DIFF
--- a/mythtv/programs/mythfrontend/playbackbox.cpp
+++ b/mythtv/programs/mythfrontend/playbackbox.cpp
@@ -5307,7 +5307,7 @@ void RecMetadataEdit::QueryComplete(MetadataLookup *lookup)
     m_episodeSpin->SetValue(lookup->GetEpisode());
     if (!lookup->GetSubtitle().isEmpty())
     {
-        m_descriptionEdit->SetText(lookup->GetSubtitle());
+        m_subtitleEdit->SetText(lookup->GetSubtitle());
     }
     if (!lookup->GetDescription().isEmpty())
     {


rcrdnalor:
Interesting. This bug exists from the first commit 2014 until now. So it is a 'stable' bug ;-).

PR #671, opened 2022-11-29
MythContext refactoring (remove gContext)

ulmus-scott:
I had planned to do more, but I got fed up with the use of global state and Singletons, so this is all I have done for now.

##### Checklist


ulmus-scott:
I plan on doing more later, but this stands on its own and I want to work on other things first.

linuxdude42:
I like the idea, but there appear to be multiple orthogonal sets of changes here.

1) Removing gcontext
2) Moving generate_file_url function
3) Moving constants from backendselect to ssdp.h and configuration.h
4) Moving functions from .h to .cpp files
5) Removal of some cleanup code. Part of those functions is related to the  context change, but there's also pidfile cleanup. Is that no longer necessary either?
6) Changes to the imagemetadata functions.

This should be split into multiple pull requests.

ulmus-scott:
> I like the idea, but there appear to be multiple orthogonal sets of changes here.
> 
>     1. Removing gcontext

This was the main goal at this time.

> 
>     2. Moving generate_file_url function

OK, this is probably independent, but I found it while removing mythcorecontext.h from header files.

> 
>     3. Moving constants from backendselect to ssdp.h and configuration.h

Moving the constants to configuration.h was necessary to move the DatabaseParamCache into MythDB.  Moving to ssdp.h was just to put the remaining constant in a better place.

> 
>     4. Moving functions from .h to .cpp files

I got tired of having to recompile basically every file unnecessarily due to transitive includes from header files.  Except for removing mythcontext.h, I think this can be split off.

> 
>     5. Removal of some cleanup code. Part of those functions is related to the  context change, but there's also pidfile cleanup. Is that no longer necessary either?

pidfile is only used by mythbackend.  I saw it was unused so I removed its unused instances.  The command line parsers need additional modification to remove the unused option, however.

> 
>     6. Changes to the imagemetadata functions.
> 

I must have noticed that when removing mythcorecontext.h, but it should be independent.

> 
> This should be split into multiple pull requests.

I'll reorder and open more pull requests.

PR #685, opened 2022-12-27
Fix big endian linking issues

KungFuJesus:
This code would compile but due to the way C++ handles templated types, it would look for this inlined routine to be exported at link time since the only templated types defined are Qt flavored integers.

Ideally this is generating single instruction byteswaps similar to what happens when you use compiler builtins but admittedly I never checked.

##### Checklist

- [ ] contribution is in a branch rebased against [master](https://github.com/MythTV/mythtv)


kmdewaal:
I do appreciate the effort spent in fixing this, but it looks like this has already been fixed in master in commit 8960641b32f751b8198ace9bbfc3378dac20c061 on March 23. Considering that it apparently also needs fixing in version 32, I think it is best to cherry-pick that commit into version 32.

KungFuJesus:
Works for me

kmdewaal:
Fixed with commit cb65c9344d3b72ff1d55e536eb748b994583382c .

PR #701, opened 2023-01-24
Match by date when no time specified in TvMaze DB

### LINKED ISSUE #654: Tvmaze metadata lookup failures

SteveErl:
Sometimes, the TvMaze database only specifies a date and not a time for an episode or set of episodes. This is most common in webChannel originated shows. In such a case, the 'Find Episode by Timestamp' syntax usually fails to find an exact or close match.
For example:
    `tvmaze.py -N "Criminal Minds" "2022-11-24 21:00:00"`

For this episode, in the TvMaze database
 ```
    airtime =
    airdate = 2022-11-24
```
In this code update, we detect when no **airtime** is specified in the database, and apply a match-by-date behavior.

The match-by-date results are only used as a last resort. First we select exact timestamp matches. If there are none of those, we select close timestamp matches. If there are none of those, then we'll select date only matches.

For example:
```
tvmaze.py -N "Criminal Minds" "2022-11-24 21:00:00"
<metadata>
  <item>
    <title>Criminal Minds</title>
    <subtitle>Just Getting Started</subtitle>
    ...
    <season>16</season>
    <episode>1</episode>
    <inetref>81</inetref>
    ...
  </item>
  <item>
    <title>Criminal Minds</title>
    <subtitle>Sicarius</subtitle>
    ...
    <season>16</season>
    <episode>2</episode>
    <inetref>81</inetref>
    ...
  </item>
</metadata>
```
Note that there were 2 episodes released on 2022-11-24, so both episodes match-by-date and appear in the above results.

In addition, a display of **runtime** has been added for the -M invocation option. For example
```
tvmaze.py -M "Fire Country"
<metadata>
  <item>
    <title>Fire Country</title>
    ...
    <inetref>60339</inetref>
    <collectionref>60339</collectionref>
    <language>en</language>
    <releasedate>2022-10-07</releasedate>
    <userrating>6.300000</userrating>
    <popularity>99.0</popularity>
    <year>2022</year>
    <runtime>60</runtime>
    ...
  </item>
</metadata>
```
Resolves #654

##### Checklist

### DIFF
--- a/mythtv/programs/scripts/metadata/Television/tvmaze.py
+++ b/mythtv/programs/scripts/metadata/Television/tvmaze.py
@@ -98,6 +98,7 @@ def buildList(tvtitle, opts):
         m.collectionref = check_item(m, ("collectionref", str(show_info.id)), ignore=False)
         m.language = check_item(m, ("language", str(locales.Language.getstored(show_info.language))))
         m.userrating = check_item(m, ("userrating", show_info.rating['average']))
+        m.runtime = check_item(m, ("runtime", show_info.runtime))
         try:
             m.popularity = check_item(m, ("popularity", float(show_info.weight)), ignore=False)
         except (TypeError, ValueError):
@@ -247,6 +248,7 @@ def buildNumbers(args, opts):
                 episodes = []
             time_match_list = []
             early_match_list = []
+            date_match_list = []
             minTimeDelta = timedelta(minutes=60)
             for i, ep in enumerate(episodes):
                 if ep.timestamp:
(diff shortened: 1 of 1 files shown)


SteveErl:
@rcrdnalor Could you please merge this pull in? It's been pending for 3 years now, and I'd like to close #654.

Issue #700, opened 2023-01-20
Add metadata lookup by timestamp to tmdb3tv.py


SteveErl:
**Is your feature request related to a problem? Please describe.**
When using "TheMovieDB.org V3 television" as the Television Metadata Source, and recording from a Manual Record rule or an automatic record rule based off of OTA broadcast information which does not specify a Subtitle field for TV shows, specific metadata for a TV episode is not available.

**Describe the solution you'd like**
In the tvmaze.py metadata script, there is an option to find a specific TV episode by providing the name of the show, and a timestamp for the initial airing. When recording the initial airing of a TV episode, use of this option allows us to retrieve specific episode metadata information, such as Season, Episode, Subtitle, and Description.

The same capability should be added to the TMDB metadata retrieval script for TV shows. Users who have chosen to use "TheMovieDB.org V3 television" will be able to collect more complete metadata information for first broadcast recordings.

**Describe alternatives you've considered**
The only alternative is for the user to select TvMaze instead of TMDB for TV show metadata collection.

rcrdnalor:
The discussion on PR #702 led to the following open questions
which need to be resolved before MythTV can add this optional feature:

 - Exact definition of an `air_date` of an episode from given tv-series.
   The forum says in 2018, its is in UTC, which was revised by the same person 2020
   as "it's assumed to be 00:00 in the timezone of the original airing network."
   A final statement from themoviedb.org about the definition of `air_date` and
   how they handle falsely entered data is welcome.
   
 - Correctness of the data `air_date` provided by the API from themoviedb.org
   The mentioned PR found some discrepancies. This needs to be verified globally,
   not only on some examples from North America.
   If there are more discrepancies, MythTV needs to evaluate if this way of finding
   'an episode by given air_date' is of any use. 
   
 - Finally, the script `tmdb3tv.py` which may interpret the provided data falsely.
   This needs to be checked by code review and tests using other metadata provider.
 
Note:
Currently, there is no way for MythTV to differentiate between wrong and correct metadata.
In other words, once the data is entered in the mythconverg database, one cannot overwrite
it easily, only by external scripts using MythTV's Services API or the MythTV Bindings.
In general, the users are very sensitive when it comes to the handling of `air_date`:
See issues #572 and #628.

garybuhrmaster:
[...]

PR #702, opened 2023-01-24
Add metadata lookup by timestamp to tmdb3tv.py

SteveErl:
This change adds **lookup by timestamp** syntax options:

```
    tmdb3tv.py -N <title> <date time>
    tmdb3tv.py -N <inetref> <date time>
```

For example:

```
tmdb3tv.py -N "The Blacklist" "2021-01-29 19:00:00"
<metadata>
  <item>
    <title>The Blacklist</title>
    <subtitle>Elizabeth Keen</subtitle>
    <description>As Red and the task force search for Liz, she sets a new plan in motion that has catastrophic consequences.</description>
    <season>8</season>
    <episode>4</episode>
    <inetref>46952</inetref>
    <collectionref>46952</collectionref>
    <language>en</language>
    <releasedate>2021-01-29</releasedate>
    ...
  </item>
</metadata>
```
In the above example, by specifying the start of the recording time, we were able to acquire the Season, Episode, Subtitle, and Description. Without this new capability, the -N invocation would fail, and would be followed by a -M invocation, which would provide only a generic description of the TV series.

Please note that the TMDB database only stores the date, and not the time, of the initial airing. The match cannot be as precise as in a TvMaze lookup, which has access to the date, time, and timezone. If there are multiple episodes aired on a single date, the TMDB script will generate a list with all the episodes matching the date in the given timestamp.

Resolves: #700

##### Checklist

### DIFF
--- a/mythtv/bindings/python/tmdb3/tmdb3/lookup.py
+++ b/mythtv/bindings/python/tmdb3/tmdb3/lookup.py
@@ -317,6 +317,7 @@ def buildEpisode(args, opts):
     from lxml import etree
     from MythTV.tmdb3 import searchSeries
     from MythTV import datetime
+    from datetime import timedelta
 
     if query.isnumeric():
         inetref = query
@@ -346,7 +347,15 @@ def buildEpisode(args, opts):
         try:
             # The database provides air_date, but not air_time,
             # so extract the date part of the timestamp.
-            timeStampDate = datetime.strptime(subtitle, "%Y-%m-%d %H:%M:%S").date()
+            dtInLocalZone = datetime.strptime(subtitle, "%Y-%m-%d %H:%M:%S")
+            # To ensure that the episode airdates are in line with
+            # the dates used by TV guides and listings worldwide,
+            # episodes that start airing at or after midnight but
+            # before 5:00am are considered part of the previous day.
+            if dtInLocalZone.hour < 5:
+                timeStampDate = (dtInLocalZone - timedelta(days=1)).date()
+            else:
+                timeStampDate = dtInLocalZone.date()
         except ValueError:
             timeStampDate = None
     else:
--- a/mythtv/bindings/python/tmdb3/tmdb3/lookup.py
+++ b/mythtv/bindings/python/tmdb3/tmdb3/lookup.py
@@ -303,9 +303,11 @@ def buildTVList(query, opts):
                                     xml_declaration=True)
 
 #   for buildEpisode, args are one of the following
-#   - title subtitle         (-N command line)
-#   - inetref subtitle       (-N command line)
-#   - inetref season episode (-D command line)
+#   -N <title> <subtitle>            e.g. -N "The Blacklist" "Elizabeth Keen"
+#   -N <title> <date time>           e.g. -N "The Blacklist" "2021-01-29 19:00:00"
+#   -N <inetref> <subtitle>          e.g. -N 46952 "Elizabeth Keen"
+#   -N <inetref> <date time>         e.g. -N 46952 "2021-01-29 19:00:00"
+#   -D <inetref> <season> <episode>  e.g. -D 46952 8 4
 def buildEpisode(args, opts):
 
     query = args[0]
@@ -314,6 +316,7 @@ def buildEpisode(args, opts):
     from MythTV import VideoMetadata
     from lxml import etree
     from MythTV.tmdb3 import searchSeries
+    from MythTV import datetime
 
     if query.isnumeric():
         inetref = query
@@ -338,6 +341,17 @@ def buildEpisode(args, opts):
         sys.stdout.write('ERROR: Episode not found: ' + str(args))
         return 9
 
+    if subtitle:
+        # Check whether the 'subtitle' is really a timestamp.
+        try:
+            # The database provides air_date, but not air_time,
+            # so extract the date part of the timestamp.
+            timeStampDate = datetime.strptime(subtitle, "%Y-%m-%d %H:%M:%S").date()
+        except ValueError:
+            timeStampDate = None
+    else:
+        timeStampDate = None
+
     # process seasons backwards because it is more likely
     # that you have a recent one than an old one
     while season_number > 0:
@@ -347,6 +361,10 @@ def buildEpisode(args, opts):
             break
         for ep_num, ep in season.episodes.items():
             if ep.name == subtitle:
+                episode = ep
+                episode_number = int(ep_num)
+                break
+            if timeStampDate and (ep.air_date == timeStampDate):
                 episode = ep
                 episode_number = int(ep_num)
                 break


rcrdnalor (commented):
A datetime or date object without an associated timezone is ambiguous.
Please check with themoviedb.com the timezone of ''air_date' of a given episode
and report back.
If the timezone is 'UTC' than the API is not in accordance with the tvmaze grabber.

SteveErl:
@rcrdnalor : It took quite a bit of digging, but I did find confirmation that all the dates are stored in UTC. I'll update my  code to adjust the date based on the users timezone.

While trying to come up with a test case, I discovered that there's a bug in our tvmaze.py script. When doing a lookup by timestamp, it fails to get specific episode information when the timezone adjusted starting time is shortly after midnight. I'll open a new issue to fix this bug.

SteveErl:
@rcrdnalor : The documentation for TMDB3 is incredibly vague and inconsistent about timezones. There is only one API command where they let you specify a timezone. The **_GET /tv/airing_today_** command. In the API documentation for that command, they say the query defaults to the US Eastern Time zone (UTC-05:00), a.k.a America/New_York.

In 2018, one user asked about UTC:
https://www.themoviedb.org/talk/5b5793520e0a2673dc0402cb
**Question**: All dates ** release_date, first_air_date, last_air_date, air_date** tmdb provides in UTC. Am I right?
**Staff Response**: You are correct, all UTC.

But, this isn't true, and somebody pointed it out in 2020.
https://www.themoviedb.org/talk/5e7f5e84a275020016eec583
**Staff Response**: It's not assumed to be UTC, it's assumed to be 00:00 in the timezone of the original airing network.

From my testing this still seems to be wrong for shows starting near midnight. I have a recording of an NBC late night talk show which starts at 11:35pm in the US Central timezone, which means it's 00:35am of the next day in the US Eastern timezone, which is considered the timezone of the original airing network. When I shift the recording time from 11:35pm to 00:35am, and lookup by date, TMDB returns the metadata for the next day's episode, which is wrong. To further complicate things, the broadcast networks re-broadcast the shows 2 hours later for the Mountain and Pacific timezones.

Due to these sorts of issues, the TvMaze API has a special care policy.
https://www.tvmaze.com/faq/15/episodes
`The time should be in the local timezone where the episode premiered. If the show's network operates in multiple timezones, the earliest timezone should be used; for example in the United States this is Eastern time. Special care should be taken when episodes air at or after midnight. To ensure that the episode airdates on TVmaze are in line with the dates used by TV guides and listings worldwide, episodes that start airing at or after midnight but before 5:00 are considered part of the previous day. For example, if an episode airs after midnight between Monday the 1st and Tuesday the 2nd, the airdate should be set to the 1st; even though technically the proper date would be the 2nd. (API users should take note that the property "airdate" uses the date as described here; while "airstamp" uses the technically correct date).`

The bug in tvmaze.py, I mentioned in the previous comment, was that it didn't account for this special care policy. I'm tempted to apply the same fix to the TMDB3 script, even though they never specify that they are following the same sort of policy.

SteveErl:
I haven't mixed anything up. I had noticed that tmdb3.py is susceptible to errors when the starting time of a show slips past midnight. Before my latest commit:
```
$ TZ=America/Chicago /usr/share/mythtv/metadata/Television/tmdb3tv.py -l en -a US -N "Late Night With Seth Meyers" "2023-01-25 23:37:00"
<metadata>
  <item>
    <title>Late Night with Seth Meyers</title>
    <subtitle>Colin Quinn, Gabrielle Union</subtitle>
    <description>Colin Quinn, Gabrielle Union</description>
    <season>10</season>
    <episode>53</episode>
    <inetref>61818</inetref>
    <collectionref>61818</collectionref>
    <language>en</language>
    <releasedate>2023-01-25</releasedate>
    <runtime>41</runtime>
    <categories>
      <category name="Talk"/>
    </categories>
    <people>
      <person name="Seth Meyers" character="Self - Host" department="Actors" job="Actor" url="http://www.themoviedb.org/people/62856" thumb="http://image.tmdb.org/t/p/original/nCcvpJFHWYHfu5oPxTOyA1uSdqA.jpg"/>
    </people>
    <images>
      <image type="fanart" url="http://image.tmdb.org/t/p/original/5AkPhazx8F0Ht74CUdJU03vNzBi.jpg" thumb="http://image.tmdb.org/t/p/w300/5AkPhazx8F0Ht74CUdJU03vNzBi.jpg"/>
    </images>
  </item>
</metadata>
```

When the same show is recorded in Eastern timezone we get:

```
$ TZ=America/New_York /usr/share/mythtv/metadata/Television/tmdb3tv.py -l en -a US -N "Late Night With Seth Meyers" "2023-01-26 00:37:00"
<metadata>
  <item>
    <title>Late Night with Seth Meyers</title>
    <subtitle>Alex Wagner, Bowen Yang</subtitle>
    <description>Alex Wagner, Bowen Yang</description>
    <season>10</season>
    <episode>54</episode>
    <inetref>61818</inetref>
    <collectionref>61818</collectionref>
    <language>en</language>
    <releasedate>2023-01-26</releasedate>
    <runtime>41</runtime>
    <categories>
      <category name="Talk"/>
    </categories>
    <people>
      <person name="Seth Meyers" character="Self - Host" department="Actors" job="Actor" url="http://www.themoviedb.org/people/62856" thumb="http://image.tmdb.org/t/p/original/nCcvpJFHWYHfu5oPxTOyA1uSdqA.jpg"/>
    </people>
    <images>
      <image type="fanart" url="http://image.tmdb.org/t/p/original/5AkPhazx8F0Ht74CUdJU03vNzBi.jpg" thumb="http://image.tmdb.org/t/p/w300/5AkPhazx8F0Ht74CUdJU03vNzBi.jpg"/>
    </images>
  </item>
</metadata>
```
In this case, the correct metadata for the show was the American/Chicago result of Season 10, Episode 53 even though America/New_York is considered the timezone of the original airing network. Nowhere in the TMDB documentation do they tell you that a show starting at midnight or shortly after should be credited to the prior date, but that's what they have implemented in their database.

In thankfully clear explanation in a different TV listings API (TvMaze) they mention that they had to adjust the airdate assigning policy  "to ensure that the episodes ... (in the database) ... _**are in line with the dates used by TV guides and listings worldwide**_, episodes that start airing at or after midnight but before ... (**cutoff**)... are considered part of the previous day. The critical part of this statement is that they point out that TV guides and listings worldwide do this sort of shifting when assigning an airdate to an episode.

For TvMaze, **cutoff** is 5am. For TMDB, they don't tell us. I had one example, "Late Night With Seth Meyers", revealing that a show between midnight and 1am is assigned the previous date. It's hard to find other TV shows with new episodes starting at or after 1am, but I tracked down "Last Call with Carson Daly" which used to air at 1:35am in New York. There's only metadata for one episode (Season 13, Episode 57) in the TMDB database, but we can compare that to the same episode in the TvMaze database. Doing so shows that **cutoff** for TMDB is at least 2am.

I haven't found any other daily TV show starting between 2am and 5am to further refine the **cutoff** value for TMDB. Since the TvMaze developers are more knowledgeable than me on the subject of 'dates used by TV guides and listings worldwide', and far more  knowledgeable than the TMDB developers, I choose to follow their example. This provides the right results for all the late night shows I've tested.

rcrdnalor:
Neither the API from themoviedb.org nor the forum specify how the `air_date` shall be used.
Over the world, there is a large community of themoviedb which fills in data for current and past movies and series.
It is likely, that these data are wrong because of missing rules. 
In contrast to tvmaze, themoviedb is used worldwide from may users. TVmaze is only useful in North-America, because it offers only a limited set of data.
The search for an episode of a tv-series by given timestamp is optional in MythTV, not required.
I repeat my previous words:
In case themoviedb.org does not provide correct data for the air-date, it is better to not provide this feature ("search for air-date")
than implementing any heuristic workarounds.
I do not want to discuss this further.

PR #734, opened 2023-03-18
Fix translation of schedule filter entries

hambre:
This commit uses 'QCoreApplication::translate' to translate the entries of the schedule filter list as this allows to pass a translation context which is needed to find the translations of these entries.

##### Checklist

- [ x] code compiles successfully without errors

### DIFF
--- a/mythtv/programs/mythfrontend/scheduleeditor.cpp
+++ b/mythtv/programs/mythfrontend/scheduleeditor.cpp
@@ -2239,8 +2239,8 @@ void FilterOptMixin::Load(void)
         {
             while (query.next())
             {
-                m_descriptions << QObject::tr(query.value(1).toString()
-                                              .toUtf8().constData());
+                m_descriptions << QCoreApplication::translate("SchedFilterEditor",
+                                    query.value(1).toString().toUtf8().constData());
             }
         }
         m_loaded = true;


linuxdude42 (commented):
You don't need to switch to QCoreApplication::translatate for this. QObject::tr can also takes an optional disambiguation string.

hambre:
> You don't need to switch to QCoreApplication::translatate for this. QObject::tr can also takes an optional disambiguation string.

Hmm, the strings are defined with the QT_TRANSLATE_NOOP macro which adds a context to the strings for translation. The 'QObject::tr' function does not have a context parameter, but a disambiguation parameter. I think they are not the same, as the 'QCoreApplication::translate' function supports both. Will test if it work anyway.

hambre:
Just checked, 'QObject::tr' with disambiguation does not work.

dengelintz:
How did you conclude it doesn't work?  We have used QObject::tr() with disambiguation as needed for ages.

hambre:
> How did you conclude it doesn't work? We have used QObject::tr() with disambiguation as needed for ages.

My statement was not that disambiguation does not work in general, but in this case. As the strings are not defines with as disambiguation comment but rather with a context. Which is not the same. Instead of the change done within the patch I used for testing:
`m_descriptions << QObject::tr(query.value(1).toString().toUtf8().constData(), "SchedFilterEditor");`
as suggested. This did not result in getting the translated strings. And as explained in my last comment it is not surprising.
The source strings are defined like this:
`static QString fs0(QT_TRANSLATE_NOOP("SchedFilterEditor", "New episode"));
static QString fs1(QT_TRANSLATE_NOOP("SchedFilterEditor", "Identifiable episode"));
static QString fs2(QT_TRANSLATE_NOOP("SchedFilterEditor", "First showing"));
static QString fs3(QT_TRANSLATE_NOOP("SchedFilterEditor", "Prime time"));
static QString fs4(QT_TRANSLATE_NOOP("SchedFilterEditor", "Commercial free"));
static QString fs5(QT_TRANSLATE_NOOP("SchedFilterEditor", "High definition"));
static QString fs6(QT_TRANSLATE_NOOP("SchedFilterEditor", "This episode"));
static QString fs7(QT_TRANSLATE_NOOP("SchedFilterEditor", "This series"));
static QString fs8(QT_TRANSLATE_NOOP("SchedFilterEditor", "This time"));
static QString fs9(QT_TRANSLATE_NOOP("SchedFilterEditor", "This day and time"));
static QString fs10(QT_TRANSLATE_NOOP("SchedFilterEditor", "This channel"));
static QString fs11(QT_TRANSLATE_NOOP("SchedFilterEditor", "No episodes"));`
As explained above, the macro adds a context to the translated string and not an disambiguation comment. (https://doc.qt.io/qt-5/qtglobal.html#QT_TRANSLATE_NOOP)

linuxdude42:
What happens if you change the context? Those strings aren't referenced from the SchedFilterEditor class but are referenced from the FilterOptMixin class. The SchedFilterEditor class doesn't do any translations.

My guess is that filtering was originally part of the  SchedFilterEditor class, and then later broken out into its own class. The context strings weren't updated at that time, so things broke.

hambre:
Yes that is what I assume has happened as well.

Changing the context would mean updating all translation files as well, otherwise the translated strings will not be found.
Is that what you want me to do?

PR #735, opened 2023-03-18
Fix detection of optical drives

hambre:
Checking for mount points is not reliable to determine if a block device is an optical drive. This commit uses the 'MediaCompatibility' property of the 'org.freedesktop.UDisks2.Drive' DBUS interface to check if the drive supports optical media.
Also changes the function name in the log messages to match the function they are called from.

##### Checklist

### DIFF
--- a/mythtv/libs/libmyth/mediamonitor-unix.cpp
+++ b/mythtv/libs/libmyth/mediamonitor-unix.cpp
@@ -191,11 +191,7 @@ static bool DetectDevice(const QDBusObjectPath& entry, MythUdisksDevice& device,
 {
     QDBusInterface block(UDISKS2_SVC, entry.path(), UDISKS2_SVC_BLOCK,
                          QDBusConnection::systemBus() );
-#if 0
-    QString devraw = block.property("Device").toString();
-    LOG(VB_MEDIA, LOG_DEBUG, LOC +
-        "DetectDevice: Raw Device found: " + devraw);
-#endif
+
     if (!block.property("HintSystem").toBool() &&
         !block.property("HintIgnore").toBool())
     {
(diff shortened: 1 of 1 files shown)


mythtv/libs/libmyth/mediamonitor-unix.cpp
```diff
@@ -229,17 +230,19 @@ static bool DetectDevice(const QDBusObjectPath& entry, MythUdisksDevice& device,
...
                 .arg(DriveProperty(drivePath, "Optical").toString()));
-        LOG(VB_MEDIA, LOG_DEBUG, LOC + QString("CheckMountable:Drive:OpticalBlank : %1")
+        LOG(VB_MEDIA, LOG_DEBUG, LOC + QString("DetectDevice:Drive:OpticalBlank : %1")
                 .arg(DriveProperty(drivePath, "OpticalBlank").toString()));
 #endif
 
         if (DriveProperty(drivePath, "Removable").toBool())
         {
-            if (readonly && !isfsmountable)
+            if (readonly && isOptical)
```
linuxdude42, line 245: Isn't the isOptical check sufficient by itself? For example, if a DVD+RW was inserted in the drive I would expect readonly to be false but this clause should still match.

linuxdude42 (commented):
I like the changes overall.

mythtv/libs/libmyth/mediamonitor-unix.cpp
rcrdnalor: A view line above it is explained why the code is written in this way:
https://github.com/MythTV/mythtv/blob/c166849536ca2785184b5b6a20aec6bfd3b7a099/mythtv/libs/libmyth/mediamonitor-unix.cpp#L225

You said
> Checking for mount points is not reliable to determine if a block device is an optical drive. 

Please explain in more detail.

hambre: The check I introduced does not use the 'Optical' or 'OpticalBlank' of the Drive interface. This only exposes as you correctly mentioned what kind of medium is currently inserted in the drive. These are also currently disabled in the code and not used. What was used before to determine if the drive was an optical drive was the 'ReadOnly' property of the 'Block' interface and the existence of the 'MountPoints' property of the 'FileSystem' interface. This was not reliable as the dvd drive I have was not returning an error when reading the 'MountPoints' property. It just returned an empty list. That caused the code to treat it like an UDisks2HDD device and thus ejecting from the DVD menu did not work. The new check uses the 'MediaCompatibility' property of the 'Drive' interface which returns a list of compatible media the drive supports. For my dvd drive it looks like this:  optical_cd, optical_cd_r, optical_cd_rw, optical_dvd, optical_dvd_plus_r, optical_dvd_plus_r_dl, optical_dvd_plus_rw, optical_dvd_r, optical_dvd_ram, optical_dvd_rw, optical_mrw, optical_mrw_w. So checking if any kind of "optical*' media is supported by the drive seems like a better test.

hambre: > Isn't the isOptical check sufficient by itself? For example, if a DVD+RW was inserted in the drive I would expect readonly to be false but this clause should still match.

Yes it may be sufficient. I just checked the 'ReadOnly' property with a CD-RW medium and a DVD-Ram medium inserted. It still has the value of 'true'. I don't know if it ever has a different value for optical drives. I can remove, the media compatibility test should be enough.

hambre:
Thanks for merging this. Is there any chance for backporting to v33?

PR #749, opened 2023-05-09
Show "Input name" field only for analog inputs

kmdewaal:
Fix a defect in mythtv-setup found by Peter when creating the WebApp configuration utility.

### DIFF
--- a/mythtv/libs/libmythtv/v4l2util.cpp
+++ b/mythtv/libs/libmythtv/v4l2util.cpp
@@ -40,7 +40,7 @@ bool V4L2util::Open(const QString& dev_name, const QString& vbi_dev_name)
     m_fd = open(dev_name.toLatin1().constData(), O_RDWR);
     if (m_fd < 0)
     {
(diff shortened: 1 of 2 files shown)


bennettpeter:
I did not include support for v4l2 devices in the web-app setup.  I have a v4l2 capture device that outputs stereo audio and either YUYV or MJPG format video. I could not get it to work with MythTV using the existing setup. I was unable to get any recording to work. The backend attempted to create a "nuv" video file but failed with many errors. I think there was also a segfault that occurred.

Do you have any evidence that v4l2 works with MythTV?

I am now using the capture device with MythTV using an external recorder and ffmpeg.

kmdewaal:
It has been years ago that I used analog inputs so no, I do not know if it still works. I always tried never to break anything but I am not really sure. I think it is OK that you leave it out of the webapp.
I think the webapp should first focus on the functionality that everybody uses and leave out the corner cases and the exotics for now.

PR #751, opened 2023-05-16
MythMusic: add percent indicator to WaveForm and shrink and dim the font

twitham1:
as the white was too bright.  Add 5 second gap to radio streams.

Also eliminate all #defines by moving the settings to class members.  Few other code cleanups that change no functionality.

##### Checklist

### DIFF
--- a/mythplugins/mythmusic/mythmusic/visualize.cpp
+++ b/mythplugins/mythmusic/mythmusic/visualize.cpp
@@ -556,7 +556,7 @@ void WaveForm::saveload(MusicMetadata *meta)
 {
     QString cache = GetConfDir() + "/MythMusic/WaveForm";
     QString filename;
-    int stream = gPlayer->getPlayMode() == MusicPlayer::PLAYMODE_RADIO;
+    m_stream = gPlayer->getPlayMode() == MusicPlayer::PLAYMODE_RADIO;
     if (m_currentMetadata)     // cache work in progress for next time
     {
         QDir dir(cache);
@@ -565,7 +565,7 @@ void WaveForm::saveload(MusicMetadata *meta)
             dir.mkpath(cache);
         }
         filename = QString("%1/%2.png").arg(cache)
-            .arg(stream ? 0 : m_currentMetadata->ID());
+            .arg(m_stream ? 0 : m_currentMetadata->ID());
         LOG(VB_PLAYBACK, LOG_INFO, QString("WF saving to %1").arg(filename));
         if (!s_image.save(filename))
         {
@@ -576,7 +576,7 @@ void WaveForm::saveload(MusicMetadata *meta)
     m_currentMetadata = meta;
     if (meta)                   // load previous work from cache
     {
-        filename = QString("%1/%2.png").arg(cache).arg(stream ? 0 : meta->ID());
+        filename = QString("%1/%2.png").arg(cache).arg(m_stream ? 0 : meta->ID());
         LOG(VB_PLAYBACK, LOG_INFO, QString("WF loading from %1").arg(filename));
         if (!s_image.load(filename))
         {
@@ -585,7 +585,7 @@ void WaveForm::saveload(MusicMetadata *meta)
         }
         // 60 seconds skips pixels with < 44100 streams like 22050,
         // but this is now compensated for by drawing wider "pixels"
-        m_duration = stream ? 60000 : meta->Length().count(); // millisecs
+        m_duration = m_stream ? 60000 : meta->Length().count(); // millisecs
     }
     if (s_image.isNull())
     {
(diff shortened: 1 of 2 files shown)


mythplugins/mythmusic/mythmusic/visualize.cpp
```diff
@@ -724,37 +734,39 @@ bool WaveForm::draw( QPainter *p, const QColor &back )
...
-    p->setPen(Qt::yellow);
-    unsigned int x = m_size.width() * m_offset / m_duration;
-    p->drawLine(x, 0, x, m_size.height());
+    p->fillRect(m_size.width() * m_offset / m_duration, 0,
+                1, m_size.height(), qRgb(128, 128, 128));
 
     if (m_showtext && m_size.width() > 500) // metrics in corners
     {
-        p->setPen(Qt::white);
+        p->setPen(qRgb(128, 128, 128)); // Qt::white);
```
linuxdude42, line 750: Qt::darkGray

mythplugins/mythmusic/mythmusic/visualize.cpp
```diff
@@ -583,15 +590,17 @@ void WaveForm::saveload(MusicMetadata *meta)
...
+    // m_font.setPointSize(14);
+    m_font.setPixelSize(m_size.height() / 60); // small to be mostly unnoticed
 }
 
 unsigned long WaveForm::getDesiredSamples(void)
 {
-    return (unsigned long) WF_AUDIO_SIZE;  // Maximum we can be given
+    // could be an adjustable class member, but this hard code works well
+    // return (unsigned long) WF_AUDIO_SIZE;  // Maximum we can be given
+    return 4096;           // maximum samples per update, may get less
```
linuxdude42, line 603: Its better to have the readable name here than a magic number.  See https://clang.llvm.org/extra/clang-tidy/checks/readability/magic-numbers.html

mythplugins/mythmusic/mythmusic/visualize.h
```diff
@@ -140,10 +140,6 @@ class MonoScope : public StereoScope
 
 // WaveForm - see whole track - by twitham@sbcglobal.net, 2023/01
 
-#define WF_AUDIO_SIZE 4096     // maximum samples to process at a time
-#define WF_WIDTH 1920   // image cache size, will scale to any display
-#define WF_HEIGHT 1080
-
```
linuxdude42, line 146: These should stay.  You could convert them to something like the following so that the compiler can understand and process them instead of doing a blind text substitution.

static constexpr uint32_t WF_AUDIO_SIZE {4096};
static constexpr int32_t WF_WIDTH {1920};
static constexpr int32_t WF_HEIGHT {1080};

mythplugins/mythmusic/mythmusic/visualize.cpp
```diff
@@ -530,6 +534,9 @@ bool MonoScope::draw( QPainter *p, const QColor &back )
 ///////////////////////////////////////////////////////////////////////////////
 // WaveForm - see whole track - by twitham@sbcglobal.net, 2023/01
 
+// static class members survive size changes for continuous display
+QImage WaveForm::m_image {nullptr}; // picture of spectrogram
+
```
linuxdude42, line 539: This doesn't handle two frontends playing different songs at the same time. Maybe a hash based on the filename? You'd also need some way to clean up when the player exits.

linuxdude42 (commented):
A couple of comments.

mythplugins/mythmusic/mythmusic/visualize.cpp
twitham1: fixed, thanks!

mythplugins/mythmusic/mythmusic/visualize.cpp
twitham1: Fixed to WFAudioSize now defined in visualize.h

mythplugins/mythmusic/mythmusic/visualize.h
twitham1: WFAudioSize restored as you suggest.  Width and height moved a few lines lower to:

    QSize         m_wfsize {1920, 1080}; // picture size

Are you saying the define or static constexpr is preferred over this?  Even if used only in this definition?  The .cpp then uses m_wfsize.height() and m_wfsize.width(), getting those values from this QSize rather than the former fixed numbers.  Perhaps reverting to static numbers is more efficient than QSize?

mythplugins/mythmusic/mythmusic/visualize.cpp
twitham1: Two frontends are in 2 processes so they have their own independent copies.  This 1 location per process would be an issue only if 1 frontend is displaying 2 songs, right?  I think this is currently impossible.

But since I already save/restore from disk I reverted it back to being destroyed and recreated on every resize.  This reverts to the noticeable blank and redraw time that I was eliminating by keeping the 1 image per process.

mythplugins/mythmusic/mythmusic/visualize.h
linuxdude42: static const is preferred.  I accept that 1920x1080 is a common enough size that someone reading the code would recognize it, so I won't say anything more about that.

I don't have any problem with using a QSize object.

mythplugins/mythmusic/mythmusic/visualize.cpp
linuxdude42: I think I keyed off the word "static" in the comment. If this is a non-static class member then its one per object and my original comment is wrong.

linuxdude42: I looked again now that I have more time then this morning. You changed WaveForm::m_image to a static, which means its shared by all instances of the object. You're right though, the there's probably only one object instance per frontend process, and only one frontend process per machine, so in reality it probably doesn't matter.

twitham1: Excellent, I have restored the static background image with an appropriate s_ prefix rather than m_.  A new comment documents how this works for now and could change back to class member if the object quits getting destroyed at resize or more than 1 needs displayed in the future.

PR #752, opened 2023-05-16
MythMusic: Spectrogram (and SpectrumDetail) visualization

twitham1:
I am pleased to offer a new Spectrogram visualization for Myth Music!

My custom FFT window is optimized for both frequency and time to show all frequencies quickly.  The Mel Scale distributes audible frequencies evenly across the display space in highest resolution.

I experimented with a colored graph but detail is harder to see in color.  Gray scale intensities more clearly show the sound details.  So now the color shows only the current spectrum as it draws the gray graph.  This makes it easy to see the high amplitudes drawing the brightest frequency dots.

**2023/06/13:** added a second visual SpectrumDetail which is simply the color spectrum turned sideways, like Spectrum, with no historical graphing.

##### Checklist

### DIFF
--- a/mythplugins/mythmusic/mythmusic/visualize.cpp
+++ b/mythplugins/mythmusic/mythmusic/visualize.cpp
@@ -998,23 +998,24 @@ bool Spectrogram::processUndisplayed(VisualNode *node)
         int count = 0;
         for (auto j = prev + 1; j <= index; j++)
         {    // for the freqency bins of this pixel, find peak or mean
-
-            // // linear magnitude:
-            // tmp = sqrt(sq(m_dftL[2 * j]) + sq(m_dftL[2 * j + 1]));
-
-            // power spectrum (dBm):
-            tmp = 10 * log10(sq(m_dftL[2 * j]) + sq(m_dftL[2 * j + 1]));
+            tmp = sq(m_dftL[2 * j]) + sq(m_dftL[2 * j + 1]);
             left  = m_binpeak ? (tmp > left  ? tmp : left ) : left  + tmp;
-            tmp = 10 * log10(sq(m_dftR[2 * j]) + sq(m_dftR[2 * j + 1]));
+            tmp = sq(m_dftR[2 * j]) + sq(m_dftR[2 * j + 1]);
             right = m_binpeak ? (tmp > right ? tmp : right) : right + tmp;
             count++;
         }
         if (!m_binpeak)
-        {			// mean of the frequency bins
+        {                       // mean of the frequency bins
             (count > 0) || (count = 1);
             left /= count;
             right /= count;
         }
+        // linear magnitude:           sqrt(sq(real) + sq(im));
+        // left = sqrt(left);
+        // right = sqrt(right);
+        // power spectrum (dBm): 10 * log10(sq(real) + sq(im));
+        left = 10 * log10(left);
+        right = 10 * log10(right);
 
         // float bw = 1. / (16384. / 44100.);
         // float freq = bw * index;
--- a/mythplugins/mythmusic/mythmusic/visualize.cpp
+++ b/mythplugins/mythmusic/mythmusic/visualize.cpp
@@ -864,20 +864,34 @@ static class WaveFormFactory : public VisFactory
 QImage Spectrogram::s_image {nullptr}; // picture of spectrogram
 int    Spectrogram::s_offset {0};      // position on screen
 
-Spectrogram::Spectrogram()
+Spectrogram::Spectrogram(bool hist)
 {
-    LOG(VB_GENERAL, LOG_INFO, QString("Spectrogram : Being Initialised"));
+    LOG(VB_GENERAL, LOG_INFO,
+        QString("Spectrogram : Being Initialised, history=%1").arg(hist));
 
-    m_fps = 15;                 // needed?  right?
+    m_history = hist;   // historical spectrogram?, else spectrum only
+
+    m_fps = 40;         // getting 1152 samples / 44100 = 38.28125 fps
+
+    if (s_image.isNull())  // static histogram survives resize/restart
+    {
+        s_image = QImage(
+            m_sgsize.width(), m_sgsize.height(), QImage::Format_RGB32);
+    }
+    m_image = m_history ? &s_image : new QImage(
+        m_sgsize.width(), m_sgsize.height(), QImage::Format_RGB32);
 
     m_dftL = static_cast<FFTSample*>(av_malloc(sizeof(FFTSample) * m_fftlen));
     m_dftR = static_cast<FFTSample*>(av_malloc(sizeof(FFTSample) * m_fftlen));
 
     m_rdftContext = av_rdft_init(std::log2(m_fftlen), DFT_R2C);
 
-    m_scale.setMax(m_fftlen / 2, m_sgsize.height() / 2);
+    m_scale.setMax(m_fftlen / 2, m_history ?
+                   m_sgsize.height() / 2 : m_sgsize.width());
     m_sigL.resize(m_fftlen);
     m_sigR.resize(m_fftlen);
+
+    // TODO: promote this to a separate ColorSpectrum class
 
     // cache a continuous color spectrum ([0-1535] = 256 colors * 6 ramps):
     // 0000ff blue              from here, G of RGB ramps UP to:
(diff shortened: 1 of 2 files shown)


mythplugins/mythmusic/mythmusic/visualize.cpp
```diff
@@ -937,7 +1175,7 @@ bool Spectrum::process(VisualNode *node)
     for (auto k = i; k < FFTW_N; k++)
     {
         m_dftL[k] = (FFTComplex){ .re = 0, .im = 0 };
-        m_dftL[k] = (FFTComplex){ .re = 0, .im = 0 };
+        m_dftR[k] = (FFTComplex){ .re = 0, .im = 0 };
     }
```
twitham1, line 1179: I started to fix bugs in Spectrum based on what I learned in Spectrogram.  I'm sure that Spectrum is not right since it is not processing the whole signal.  But I will submit those fixes in a separate pull request.

Jpilk:
I have this working, built as mythtv-plugins-34.Pre.207.ga8223da706-100.fc37.x86_64.rpm

It shows the musical structures well.  Thanks!

twitham1:
> I have this working, built as mythtv-plugins-34.Pre.207.ga8223da706-100.fc37.x86_64.rpm
> 
> It shows the musical structures well. Thanks!

You're welcome!  Today's commit adds option to plot peak or mean.  If I could get a KeyPress action, then I would let a button on the remote toggle this.  But the handleKeyPress appears to never get called by frontend.

Jpilk:
OK, it's looking good.  A few comments on visualize.cpp

Block at line 990.  I think it might be better to have the freq-bin boundaries in geometric progression.  j = j * factor rather than j++

Block at line 1006.  Work in linear power.  Postpone the 10 * log10 dBM conversion to after line 1015.  

Line 1200 s/dftL/dftR

Jpilk:
Screenshot_20230525_130724

twitham1:
The Y axis is currently logarithmic, a common spectrogram display.  This is due to prev and index coming from the LogScale m_scale, set at the end of the loop for the next pixel.  The FFT returns data every ~2.7Hz for a 44100Hz song.  So at higher frequencies, j is combining up to 60+ bin results into that single pixel.  So j should not skip any of that valid data for that pixel.  But perhaps you were thinking of a geometric Y axis instead?

You're right that 10 * log10 can be done outside the j loop.  Thanks and done!

Line 1200 is beyond my code and in the original Spectrum.  I'd like to get this merged first then fix Spectrum in a separate pull request.  In fact, Spectrum is wrong as it doesn't get the full signal so it can't get the frequencies right.  Rather than fix it, we could use mine instead.  Imagine my colored section rotating 90 degrees then filling the whole screen like Spectrum.  It would be a simple subclass.  Squares would need refactored too.

If folks like the original Spectrum with the wide bars and decay, then we can both fix it and add mine as a new SpectrumDetail or similar.

Jpilk:
The current display looks fine to me, but maybe in principle it should show the total power per bin rather than the mean power per fft element within each bin?    Lines  1010 and 1011?

That would increase the densites in the upper portion of each plot.

Jpilk:
I now have a build in which the Left channel display uses the 'total power' in each bin and the Right channel display uses the 'mean power in each FFT item within a bin'.  As expected, the Total Power display  (by commenting out the 'left /=  count' line ) gives more weight to the higher frequencies, but the visual effects are much the same.

twitham1:
I'd like the 1 key to cycle the left channel through mean, peak and total while the 3 key does the same for the right channel making it easy to compare them as you did.  The 2 key would cycle them both.  Current values and key clues could be displayed briefly.  But this can come later when I figure out how to get the frontend to call the handleKey of the visuals.

Jpilk:
I was just thinking of opening a new issue related to visualisation when I came across this: 

Visualisation is disabled in playback of DVB radio if 'Interactive TV' aka DSMCC is enabled at Frontend Setup > Video > Playback OSD.    That screen also has a 'Preferred Visualiser' selector that doesn't show Spectrogram.

DSMCC shows (in some channels) info about the current programme content.   It also affects operation of the cutlist editor.

But in playback with DSMCC disabled, the Menu key > Audio > Visualisation offers a list of visualisers; they work but the list does not include Spectrogram.   This was my 'new' feature, and probably follows from the missing preference option.  Could it be related to 'your' key-blindness?

Spectrogram is offered and does work in 'Play Music' and 'Play Radio Stream'.  

I would like a more convenient way to control DSMCC status.

Jpilk:
In fact, I find that when watching a normal TV recording

M > Audio > Visualisation

gives  a list of visualisers, which does not include Spectrogram.  Four will share the full-screen display  with the video.  Goom and Goom HD  take over.

Jpilk:
I asked 'users' about these lists of visualisers.  Here's a link:

http://lists.mythtv.org/pipermail/mythtv-users/2023-June/411894.html

twitham1:
I have only ever used the visualizers in MythMusic.   Both Play Music and Radio Stream share the same visualizers.  I don't have any audio-only TV channels to try visualizers there.  I don't get an M > Audio > Visualization option.  But I see that David clarified that is separate code in a different path.  So porting this (and many other visualizers) to there or merging them to one code base would be a much larger effort.

Jpilk:
Sure.  It looks good as it is now.  Thanks.

Jpilk:
https://github.com/MythTV/mythtv/pull/752#issuecomment-1574388709

FWIW the visualisers offered for non-mythmusic files are defined in /mythtv/libs/libmythtv/visualisations

and /mythtv/libs/libmythtv/mythplayervisualiserui.cpp, .h are probably involved too.

twitham1:
> I didn't notice the Y-axis scaling. Sorry.

I had simply used the LogScale from Spectrum.  This certainly looks good enough for this initial version.

But I had an idea to label notes, which is impossible on a linear or even log scale.  We would need to zoom in further on the low notes to make them even wider.  A Mel scale could help - it is more linear in low frequencies then transitions to log scale higher.  But now I stumbled upon the possibly even better Constant-Q Transform or CQT.  Run this to preview what that might look like:

    ffplay -f lavfi 'amovie=file.mp3, asplit [a][out1]; [a] showcqt [out0]'

(from ffmpeg-filters(1) manual)

So this is more complex to do.  Even if I do this, I'd like log, and maybe even linear, to be options.  This could be done in separate pull request later.

Jpilk:
I haven't looked at the current details of the Y scale. What I see now seems visually ok and I don't know what case there might be for a more rigorously defined display.  Perhaps Mozart had one for the Allegri Miserere.

And the fft is ... fast.   On my box now I typically get a cpu usage around 45%

twitham1:
Yes, FFT is excellent.  My i3-550 from 2010 doesn't even use a whole core to run this at 40 fps.  But ffplay showcqt is also fast.  It might run a separate FFT per octave, so not much more compute than one FFT over whole range.

But switching to Mel Scale worked well here.  It now zooms in on the frequencies that we hear while still covering the whole 22kHz.  There is code that can be uncommented in Spectrogram::process if you would like to see the pixel, FFT bin and frequency.

Jpilk:
I wasn't aware of the Mel frequency mapping.   It looks ok here.  I still think it would be better to use 'total power in bin' rather than mean or peak fft values, though.  

atop now reports 90-ish % cpu load, maybe higher for SpectrumDetail.

nvidia OSD shows GL 20 FPS, higher than before.

The '6' key cycles the visualiser in use, set in the key editor for Music.  Perhaps you could add options there?

twitham1:
I agree, total power sounds reasonable.  I cannot find it now, but I had read an article that said when compressing multiple FFT bins into a display pixel, peak is most often used for display and mean for further calculations.  Total wasn't even mentioned!  After trying it I think I understand why.  Total appears more "noisy" meaning it is more white and less black.  I even get quick "lightning strikes" all the way to the 22khz edge that are not there using peak.  I believe this is due to the spectral leakage.  If we total, then we are adding the leaks around the detected frequency to its peak, making it taller than it really is.  Mean will instead use the leaks to pull the peak down, making the result less than the peak.  So for me, peak is the best display, which agrees with the article.

The Mel scale simply replaced the lookup array, so it has no effect on CPU usage for me.  Perhaps you are comparing the original 15 fps with the new 40 fps?  When I added SpectrumDetail, 15 fps was too slow, it looked bad.  The FFT is naturally updating over 38 fps so 40 fps is optimum update rate.

I'm not familiar with nvidia OSD.  But my change from 15fps to 40fps in the source was noticeable.

I can live without any keyboard options.  But I just reported #766 in case anyone has any ideas.  I'd rather not clutter the Music key map with something that applies to only certain visual, but maybe that is an option.

Jpilk:
OK, thanks for those comments.  It works out-of-the-box for me and adds interest.  :-)

Jpilk:
My grasp of git is shaky, and I have to report that my recent efforts have created patches that fail to apply to current master.  Perhaps a consolidation would help...

twitham1:
Well, #771 is all 4 visuals consolidated and should work against master.  The owners can opt to merge that and cancel this if they wish.  But I assumed the preference is to merge this as 2 new visuals, then merge the other as improving 2 existing visuals.

Jpilk:
Yes, I see those new features. The SpectrumDetail is striking, and similar in concept to the older Piano(Keys) display, although the signal processing is quite different.  But my preference is for the Spectrogram, showing actual audio history, or the older Goom for a loosely-audio-linked ever-changing decorative display.  Thanks anyway.  :-)

linuxdude42:
I fixed the minor conflicts and committed these change. Please check to ensure I didn't create any problems.

PR #760, opened 2023-06-14
Disable the libva GLX backend if EGL is enabled

DarkDefender:
##### Checklist

- [ ] commits are logically organised and have [good commit messages](https://chris.beams.io/posts/git-commit)

I've left the last two items unchecked on purpose as I think this might not be the solution that will be accepted. But we will see how you guys feel.

Currently mythtv always seem to pull in the `drm` and `glx` backend for libva.
On Gentoo we have removed the glx backend for libva to simplify some packaging.
This shouldn't regress the user experience as it seems all packages using libva in Gentoo (including mythtv) also supports the EGL backend and usually prefers to use the EGL backend as well.

However there is currently no way to compile mythtv without also requiring the libva glx backend, even if it will never be used. So to fix this I created the "lazy" solution of making the GLX backend only available if EGL support is not turned on.

I'm unsure if this is an acceptable solution or not. But at least it should provide a good basis for starting a discussion.

One could instead for example add a `USING_GLX` flag to allow both backends to be available as before.
However most other projects that depend on libva in the Gentoo package manager seem to have opted to drop the GLX backend and always use the drm backend. (However this of course means that you won't be able to use it on older linux dists where EGL is not available)

### DIFF
--- a/mythtv/libs/libmythtv/libmythtv.pro
+++ b/mythtv/libs/libmythtv/libmythtv.pro
@@ -538,7 +538,7 @@ using_frontend {
         DEFINES += USING_VAAPI
         HEADERS += decoders/mythvaapicontext.h
         SOURCES += decoders/mythvaapicontext.cpp
-        LIBS    += -lva -lva-x11 -lva-glx -lva-drm
+        LIBS    += -lva -lva-drm
     }
 
     using_nvdec {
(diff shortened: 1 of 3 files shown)


ulmus-scott:
@linuxdude42, since https://github.com/MythTV/mythtv/pull/1151 has been merged, this can be closed.

linuxdude42:
Closing as per Scott's comment.

Issue #464, opened 2022-01-19
Display recordedid to allow simpler file access from the desktop


Jpilk:
**Is your feature request related to a problem? Please describe.**
Please describe the problem you are trying to solve.

http://lists.mythtv.org/pipermail/mythtv-users/2022-January/408670.html

file:///home/john/SGs/CoverArtSG1/Screenshot_20220119_220900.png

--------------
Fragment of MythCenter-wide/recordings-ui.xml

        <textarea name="channel" from="basetextarea">
            <area>30,444,245,30</area>
        </textarea>

        <textarea name="timedate" from="basetextarea">
             <area>280,444,465,30</area>
             <align>hcenter</align>
	<template>%recstarttime% - %recendtime% %recstartdate%.%startyear% c%chanid% r%recordedid%</template>
        </textarea>
    
        <textarea name="filesize_str" from="basetextarea">
            <area>750,444,165,30</area>
            <align>right</align>
        </textarea>

Jpilk:
![Screenshot_20220119_220900](https://user-images.githubusercontent.com/3685166/150229225-d21ad7e4-2484-4ab2-b3e8-07e28b9e41c3.png)

Jpilk:
I would prefer to use the new 'recordedtime' format in the middle of the template definition here;  being in the UK I'm not usually much affected by this.  The form I carried over from Git behaves inconsistently with current and earlier years.

I think the display ought to give some indication of the duration of the recording, and the recordedtime as above, and the chanid and the recordedid.    " %recstarttime% - %recendtime%" gives duration and has the advantage of being linked to the filename, but recordedid should eventually point more easily to the file in its StorageGroup

Jpilk:
Putting %recordedtime% into the template doesn't deliver.  I suppose that's because it doesn't come from programinfo.cpp ?

kmdewaal:
I think that this is not information that should be on the "Watch Recordings" page. It is not relevant for normal MythTV operation but it can be useful for debugging.

The correct place for this type of additional information, such as the database recordid, is in the "second i" window. This is the popup window with detailed information that comes when, after selecting a recording in the "Watch Recordings" window, you press "i" and then "i" again. After the first "i" there is one page of additional information and after the second "i" there is a lot more.
These popup windows contain a lot of information that is not needed for normal MythTV frontend operation but that is occasionally useful for debugging, such as the filename, the recording rule used and the tuner input used. The recordid can be added there.

Jpilk:
[...]

PR #771, opened 2023-07-07
MythMusic: Improve Spectrum and Squares

twitham1:
This applies the lessons learned in Spectrogram to the original Spectrum and Squares, greatly improving them:

* FFT of complete signal
* same FFT: 16384 bins, decaying window, peak of bins
* same Mel Scale replacing log scale (much improved resolution in low frequencies, distributes the sound better across the whole screen)
* upgrade to 128 bars for better resolution

### DIFF
--- a/mythplugins/mythmusic/mythmusic/visualize.cpp
+++ b/mythplugins/mythmusic/mythmusic/visualize.cpp
@@ -108,8 +108,8 @@ void LogScale::setMax(int maxscale, int maxrange)
     if (maxscale == 0 || maxrange == 0)
         return;
 
-    m_s = maxscale;
-    m_r = maxrange;
+    m_scale = maxscale;
+    m_range = maxrange;
 
     auto domain = (long double) maxscale;
     auto range  = (long double) maxrange;
@@ -169,8 +169,8 @@ void MelScale::setMax(int maxscale, int maxrange, int maxfreq)
     if (maxscale == 0 || maxrange == 0 || maxfreq == 0)
         return;
 
-    m_s = maxscale;
-    m_r = maxrange;
+    m_scale = maxscale;
+    m_range = maxrange;
 
     m_indices.clear();
     m_indices.resize(maxrange, 0);
--- a/mythplugins/mythmusic/mythmusic/visualize.cpp
+++ b/mythplugins/mythmusic/mythmusic/visualize.cpp
@@ -37,10 +37,6 @@
 #include "musicplayer.h"
 #include "visualize.h"
 
-static constexpr int FFTW_N { 512 };
-// static_assert(FFTW_N==SAMPLES_DEFAULT_SIZE)
-
-
 VisFactory* VisFactory::g_pVisFactories = nullptr;
 
 VisualBase::VisualBase(bool screensaverenable)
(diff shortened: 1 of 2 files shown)


Jpilk:
I have master 0279fd2 running now, with MCW1.10  locally patched to display recordedid as in issue #464.

The only difficulty seen so far is inability to quit full-screen visualiser when the ident popup is continually refreshed, as with ClassicFM streaming.  That needs ctrl/c to quit mythfrontend.

http://ice-sov.musicradio.com:80/ClassicFMMP3

mythplugins/mythmusic/mythmusic/visualize.h
```diff
@@ -199,13 +199,18 @@ class MelScale
...
+
     void setMax(int maxscale, int maxrange, int maxfreq);
     double hz2mel(double hz);
     double mel2hz(double mel);
     int operator[](int index);
 
   private:
     std::vector<int> m_indices;
+    int  m_s       {0};
+    int  m_r       {0};
```
linuxdude42, line 213: These aren't very descriptive variable names.

mythplugins/mythmusic/mythmusic/visualize.cpp
```diff
@@ -1315,19 +1314,23 @@ void Spectrum::resize(const QSize &newsize)
 
     m_size = newsize;
 
-    m_analyzerBarWidth = m_size.width() / 64;
+    m_analyzerBarWidth = m_size.width() / 128;
 
     if (m_analyzerBarWidth < 6)
         m_analyzerBarWidth = 6;
```
linuxdude42, line 1320: Possibly replace this with something like:
```
m_analyzerBarWidth =
    std::max(m_size.width() / kNumBarsWanted, kMinBarSize);
```
and define the two constants in the Spectrum class with:
``` 
static constexpr int kNumBarsWanted {128};
static constexpr int kMinBarSize {6};
```
I know there are close to 15,000 other magic numbers in the code, so I'll leave it to you whether to change this. I do think this type of change makes the code more readable.

linuxdude42 (commented):
A couple of minor comments.

twitham1:
> The only difficulty seen so far is inability to quit full-screen visualiser when the ident popup is continually refreshed, as with ClassicFM streaming. That needs ctrl/c to quit mythfrontend.
> 
> http://ice-sov.musicradio.com:80/ClassicFMMP3

Wow, that's bad!  But totally unrelated to this PR.  I think I can fix it but let's work on that over on #790.

mythplugins/mythmusic/mythmusic/visualize.h
twitham1: Fixed!

mythplugins/mythmusic/mythmusic/visualize.cpp
twitham1: Personally, for a number used in 1 place, the number is easier to read than words.  Because if it is words I have to go lookup where the words resolve to a number.  So reading 128 and 6 right here is easier for me.

But certainly numbers used in more than 1 place should be words instead.

linuxdude42: You understand the code, so its not surprising that you find the numbers easier. For someone reading the code for the first time, the descriptive constant names would be easier.

PR #779, opened 2023-08-05
MythMusic: let music controls work during trackinfo popup

### LINKED ISSUE #780: MythMusic: controls fail during Info popup over Fullscreen Visualization
* **Platform**: Ubuntu 20.04

* **Package version**:

* **Component**: mythmusic

Play music, Menu -> Fullscreen Visual -> I key for Info popup.  Now hit Z for next track, Q for previous, FFWD, REW and so on.  

always

expect all controls to work

### What do you see instead?
No controls work.  Only INFO or ESCAPE work to dismiss the popup.  Also, it would be great if hitting I a second time showed the full track details.

### Additional information
I have it fixed in #779.

### LINKED ISSUE #781: MythMusic: trackinfo_popup has no access to progressbar, time, etc.
https://www.mythtv.org/wiki/Music-ui.xml#The_.22trackinfo_popup.22_window implies that trackinfo_popup can access player widgets like progressbar and time.  But this is not correct, it can use only this track and next.

* **Platform**: Ubuntu 20.04

* **Package version**:

* **Component**: mythmusic

Use progressbar or time in trackinfo_popup of music-ui.xml

always

progressbar and time widgets to display track playing position

### What do you see instead?
these widgets are unavailable.  I'd like progressbar and time to see the current playing position.

### Additional information
I have it fixed in #779.  See also related #780.

twitham1:
and let another i key press show track details and let common widgets work so theme can show for example progress bar and time.

Fixes #780, fixes #781, enhances #785, #786.

Highly recommend custom trackinfo_popup of themes add progressbar and time to assist seeking since FFWD/REW work as expected in #785.  This is provided here in the shipped defaults and MythCenter-wide.  We also add year and track number which is nice to know when skipping to next or previous track.

The XML tweaks to time and playlisttime are for #786, done here in one place to avoid conflicts.  Ideally both should be merged near the same time.

##### Checklist

- [x] (for this PR): adjust shipped themes to include progressbar and time in trackinfo_popup

### DIFF
--- a/mythplugins/mythmusic/mythmusic/visualizerview.cpp
+++ b/mythplugins/mythmusic/mythmusic/visualizerview.cpp
@@ -198,12 +198,12 @@ bool TrackInfoPopup::keyPressEvent(QKeyEvent *event)
     for (int i = 0; i < actions.size() && !handled; i++)
     {
         QString action = actions[i];
+	handled = true;
 
-        if (action == "INFO")
-        {
-            showTrackInfo(gPlayer->getCurrentMetadata());
-            handled = true;
-        }
+	if (action == "ESCAPE")
+	    Close();
+        else if (action == "INFO")
+	    showTrackInfo(gPlayer->getCurrentMetadata());
         else
             handled = false;
     }
--- a/mythplugins/mythmusic/mythmusic/visualizerview.cpp
+++ b/mythplugins/mythmusic/mythmusic/visualizerview.cpp
@@ -198,12 +198,12 @@ bool TrackInfoPopup::keyPressEvent(QKeyEvent *event)
     for (int i = 0; i < actions.size() && !handled; i++)
     {
         QString action = actions[i];
-	handled = true;
+        handled = true;
 
-	if (action == "ESCAPE")
-	    Close();
+        if (action == "ESCAPE")
+            Close();
         else if (action == "INFO")
-	    showTrackInfo(gPlayer->getCurrentMetadata());
+            showTrackInfo(gPlayer->getCurrentMetadata());
         else
             handled = false;
     }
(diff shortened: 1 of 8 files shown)


twitham1:
Hi Paul,

I believe you are thinking of trackdetail_popup.  It shows the selected track detail like you say and this PR is not touching that window.

But trackinfo_popup is available on I only over the full screen visualizer which is available only for the playing track.  What I am trying to do is be able to control the music while this is displayed.  I want to hit I again to show the more detailed info you mentioned, Z to move along to the next track and so on.  And especially since #785 fixes REW and FFWD, I want to see where I am located in the track, just like we have over in TV when we hit I and get a progress bar.  Here is what it looks like in Mythbuntu for example:

foo

With this patch, I can control the music while seeing its position without leaving the full screen visualizer.

And as I mentioned, the manual already says this should be possible so I believe it was an accident that this was not possible.

paul-h:
Ah you are right I was thinking of the other info popup I didn't realise you was talking about the one specific to the full screen visualiser.

I can't actually try it but in principle I have no objections this PR seems like a good idea. I did try to make most operations available through the menus for those users who don't have 100 buttons on their remote so in theory you should be able to control the player that way. Admittedly it's not as convenient than using dedicated key presses :)

twitham1:
Speaking of the menu, if I let M pop the menu over this temporary popup, then when this info disappears in 8 seconds, exiting the menu seg faults the frontend!  I assume this is because the parent window of the menu no longer exists.  I can workaround that by canceling the 8 second timer to make sure the parent stays around and user can dismiss it with ESCAPE.  However, the menu also fails to load the visualizer list.  I cannot figure out how this is possible.  So currently I suppress menu during info popup: user will need to ESCAPE out of INFO to use the menu just as before.

PR #785, opened 2023-08-08
MythMusic: fix REW/FFWD to seek quickly

### LINKED ISSUE #784: MythMusic: REW/FFWD plays 1 second per seek which is too slow to move long distances
As mentioned in unrelated #774, seeking via REW/FFWD plays 1 second of sound at each seek which is too slow to move long distances like several minutes in long tracks.

* **Platform**: Ubuntu 20.04

* **Component**: mythmusic

Play a music track and hit FFWD rapidly 12 times. 

always

Expect to start hearing 60 seconds later in the track immediately after FFWD is hit 12 times.  If we then REW 12 times, expect to hear near where we started from, but now rather than 12 seconds later.

### What do you see instead?

Hear 1 second of sound from 12 positions in the track before normal playback resumes at the desired minute later in the track.  Visualization does not update during the 12 seconds either.

### Additional information

It is due to maintaining the 1 second minimum buffer.  But I see a fix...

twitham1:
Fixes #784.

### DIFF
--- a/mythplugins/mythmusic/mythmusic/avfdecoder.cpp
+++ b/mythplugins/mythmusic/mythmusic/avfdecoder.cpp
@@ -458,6 +458,9 @@ void avfDecoder::run()
                 LOG(VB_GENERAL, LOG_ERR, "Error seeking");
 
             m_seekTime = -1.0;
+            // Play all pending and restart buffering, else REW/FFWD
+            // takes 1 second per keypress at the "buffered" wait below.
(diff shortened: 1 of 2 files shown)


Jpilk:
Good.  Now I see clean skips of 5 seconds forward and back on 'space' when the on-screen arrows are active,  and holding down the space key simulates FFWD/Rewind at a useful rate.   PageDown/PageUp works in full-screen visualiser mode, too.

PR #786, opened 2023-08-15
MythMusic: show playback speed only when altered

twitham1:
Playback speed can be changed faster/slower (by default W and X keys) but this is not displayed anywhere.  So this makes it hard to know the speed factor.  This simply appends the speed to time and playlisttime only when speed is changed.  A speedup of 25% looks like:

    01:36 / 03:53, 1.25

Some themes might need to tweak text area width wider or adjust align to be perfect, but it already works well with the XML in #779 and should be close in all themes.  There is no display change when playback speed is normal.

I'll update the wiki only if this is merged.

### DIFF
--- a/mythplugins/mythmusic/mythmusic/avfdecoder.cpp
+++ b/mythplugins/mythmusic/mythmusic/avfdecoder.cpp
@@ -458,6 +458,9 @@ void avfDecoder::run()
                 LOG(VB_GENERAL, LOG_ERR, "Error seeking");
 
             m_seekTime = -1.0;
+            // Play all pending and restart buffering, else REW/FFWD
+            // takes 1 second per keypress at the "buffered" wait below.
+            output()->Drain();  // (see issue #784)
         }
 
         while (!m_finish && !m_userStop && m_seekTime <= 0.0)
--- a/mythplugins/mythmusic/mythmusic/avfdecoder.cpp
+++ b/mythplugins/mythmusic/mythmusic/avfdecoder.cpp
@@ -458,9 +458,6 @@ void avfDecoder::run()
                 LOG(VB_GENERAL, LOG_ERR, "Error seeking");
 
             m_seekTime = -1.0;
-            // Play all pending and restart buffering, else REW/FFWD
-            // takes 1 second per keypress at the "buffered" wait below.
-            output()->Drain();  // (see issue #784)
         }
 
         while (!m_finish && !m_userStop && m_seekTime <= 0.0)
--- a/mythplugins/mythmusic/mythmusic/musiccommon.cpp
+++ b/mythplugins/mythmusic/mythmusic/musiccommon.cpp
@@ -2136,10 +2136,15 @@ QString MusicCommon::getTimeString(std::chrono::seconds exTime, std::chrono::sec
 {
     if (maxTime <= 0ms)
         return MythDate::formatTime(exTime,
-                              (exTime >= 1h) ? "H:mm:ss" : "mm:ss");
+                                    (exTime >= 1h) ? "H:mm:ss" : "mm:ss");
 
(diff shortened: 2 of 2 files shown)


Jpilk:
This works for me in mythCenter-wide:

[john@HPFed MCW19f-JP]$ diff -u music-base.xml music-base0.xml
--- music-base.xml      2023-08-16 16:02:27.349821513 +0100
+++ music-base0.xml     2022-01-29 22:38:31.000000000 +0000
@@ -722,7 +722,7 @@
         </progressbar>
 
         <textarea name="time" from="basetextarea">
-            <area>230,205,300,35</area>
+            <area>230,205,185,35</area>
             <font>basesmall</font>
         </textarea>

twitham1:
Good catch, thanks.  Since I was in the XML in #779 and was afraid of introducing conflicts, the tweaks for time fields are over there instead of here.

If interested, my fork has an "all-hacks" branch where I have all my pending PR running together.

Jpilk:
Thanks, I'll look at that.  I  think I had missed one set.  And I had later noticed the playlisttime truncation which you have corrected.

I'm wondering if the displayed durations ought to be the original or stretched values.  Stretched could be useful if wanting to fill a scheduling slot.;  but it might be confusing.

twitham1:
> I'm wondering if the displayed durations ought to be the original or stretched values. Stretched could be useful if wanting to fill a scheduling slot.; but it might be confusing.

I definitely want to see real track position and length, not altered.  Think of speed like a percent: 1.1 is about 10% faster so an hour of audio should take about 54 minutes.  Perhaps I should display speed factor as a percent instead so 0.85 -> 85% and 1.15 -> 115%.  I wonder which would be preferred by most?

linuxdude42:
I cherry-picked the last change and committed it.

paul-h:
> I definitely want to see real track position and length, not altered. Think of speed like a percent: 1.1 is about 10% faster so an hour of audio should take about 54 minutes. Perhaps I should display speed factor as a percent instead so 0.85 -> 85% and 1.15 -> 115%. I wonder which would be preferred by most?

My preference would be 1.1x I think we use something similar in the video player?

Of cause the proper fix would be to make it themeable now that is possible by using a template string like we do elsewhere. I'd have to look at the code to remind myself how it works but off the top of my head add a toMap() function to MusicPlayer and add keys for something like currentpos, tracklength, playbackspeed etc. Then rather than hard code that string you allow the themer to add the relevant string(s) how they want :)

twitham1:
> My preference would be 1.1x I think we use something similar in the video player?

Yes, I see the 1.1x on Menu -> Playback -> Adjust Time Stretch.  Then, like music, it is not displayed on Info.  If it were appended to the time, like in this PR, then it would automatically be on the OSD with the progress bar with no theme change needed.

For REW and FFWD, we do get 3X or 5X on the OSD.

> Of cause the proper fix would be to make it themeable now that is possible by using a template string like we do elsewhere. I'd have to look at the code to remind myself how it works but off the top of my head add a toMap() function to MusicPlayer and add keys for something like currentpos, tracklength, playbackspeed etc. Then rather than hard code that string you allow the themer to add the relevant string(s) how they want :)

In music, there is a "trackspeed" defined so I think that was exactly the intent.  But its value is not updated so the feature is not finished.  But it would be quite difficult to get all themes to decide where to put a new text area.  By appending to the time field it already works in all themes I tested and is in a consistent place.

twitham1:
Right, I noticed templates when I upgraded Mythbuntu early this year.  #793 makes the 3 components of time available to templates.

When testing, I play a podcast over an hour long and have over 100 hours in my playlist to get wider text.  Changing speed did overflow a few text areas in this case.  These were updated in #779 to show the full content.  Any truncation in other themes is a clue that speed is abnormal and a wider area is needed to show the value.

PR #791, opened 2023-08-21
MythMusic: handle streams that repeat same metadata in changing strings

### LINKED ISSUE #790: MythMusic: radio stream history can be spammed by repeated metadata
@jpilk just mentioned in unrelated #771 that some broadcasters apparently repeat identical metadata every second.  I find this unexpected (I expect the metadata only on track change) but perhaps we could handle this better.  Currently a stream like this wedges full-screen visualizer such that it can't be exited since it is constantly overlaying the trackinfo_popup.

* **Component**: mythmusic radio stream

Add a stream to this URL: http://ice-sov.musicradio.com/ClassicFMMP3 then play it.  Notice track updates every second!  Menu -> Fullscreen Visualizer.  Escape to exit fails.

always

Probably we should ignore the stream metadata if it is identical to previous data.  We should accept the data only when it changes.  That should yield just one track change notice and correct track length for a stream like this.

### What do you see instead?

Current track listed every second!

twitham1:
Fixes #790 by parsing the changed metadata string and accepting it only if title/artist/album changed.

The LOG of raw data string and its parse is moved to DEBUG level so that stations repeating the same StreamTitle in a changing string don't spam an INFO log.  Actual changes to StreamTitle are still logged to INFO as before only at first detection of the change.

### DIFF
--- a/mythplugins/mythmusic/mythmusic/avfdecoder.cpp
+++ b/mythplugins/mythmusic/mythmusic/avfdecoder.cpp
@@ -149,12 +149,13 @@ ShoutCastMetaMap ShoutCastMetaParser::parseMeta(const QString &mdata)
         auto match = rx.match(title);
         if (match.hasMatch())
         {
-            LOG(VB_PLAYBACK, LOG_INFO, QString("ShoutCast: Meta     : '%1'")
+            LOG(VB_PLAYBACK, LOG_DEBUG, QString("ShoutCast: Meta     : '%1'")
                     .arg(mdata));
-            LOG(VB_PLAYBACK, LOG_INFO,
-                QString("ShoutCast: Parsed as: '%1' by '%2'")
-                    .arg(match.captured(m_metaTitlePos),
-                         match.captured(m_metaArtistPos)));
+            LOG(VB_PLAYBACK, LOG_DEBUG,
+                QString("ShoutCast: Parsed as: '%1' by '%2' on '%3'")
+                .arg(m_metaTitlePos ? match.captured(m_metaTitlePos) : "",
+                     m_metaArtistPos ? match.captured(m_metaArtistPos) : "",
+                     m_metaAlbumPos ? match.captured(m_metaAlbumPos) : ""));
 
             if (m_metaTitlePos > 0)
                 result["title"] = match.captured(m_metaTitlePos);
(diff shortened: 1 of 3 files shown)


Jpilk:
Thanks for this.  FYI I had git problems after Gary updated his build script and my efforts have not yet improved matters; so at present I can't comment.

mythplugins/mythmusic/mythmusic/avfdecoder.cpp
```diff
@@ -553,27 +554,32 @@ void avfDecoder::checkMetatdata(void)
...
+            if (m_lastMetadataParsed != parsed)
+            {
+                m_lastMetadataParsed = parsed;
 
-            MusicMetadata mdata =  gPlayer->getDecoderHandler()->getMetadata();
-            mdata.setTitle(meta_map["title"]);
-            mdata.setArtist(meta_map["artist"]);
-            mdata.setAlbum(meta_map["album"]);
-            mdata.setLength(-1ms);
+                LOG(VB_PLAYBACK, LOG_INFO, QString("avfDecoder: shoutcast metadata changed - %1").arg(shout));
```
linuxdude42, line 571: You have two log messages, but only print them if both conditions (line 559 and line 567) are true.
Would it make sense to leave this message at line 562 (maybe at the LOG_DEBUG level), so that when debugging you can distinguish between "new" metadata arriving, and the new metadata parsing differently than before? Otherwise, this and the subsequent message seem redundant.

linuxdude42: Another argument for moving it back is that the current code prints this message before printing the two messages from parse.parseMeta.  With this change it prints after.

twitham1: Your suggestion would work for most streams.  But for a verbose one like found in this bug report, line 562 runs once per second due to the changing timestamp outside the title in the shout string.  So I can only afford to report INFO level after the parse and line 567 condition to report it only once at the title change.  I like to see the raw data and parsed result even at INFO level, as in v33.

Line 564 hands off to the parse which reports the original raw and parsed only at the DEBUG level.  This way you can see the once per second raw data in DEBUG mode for a verbose stream like this one.

Jpilk:
I'm back now with master plus these 2 PRs.  visualisers with radio streams seem fine now.  Maybe it would be better to show track info immediately on going to full-screen mode, but that might get to seem intrusive.

twitham1:
> Maybe it would be better to show track info immediately on going to full-screen mode, but that might get to seem intrusive.

This might be optional since we just came from a screen that showed what is playing.  User can hit INFO anytime he wants a reminder.

But I have thought that hitting keys like NEXTTRACK, FFWD or SPEEDUP should probably show the trackinfo so that you can see what you are doing.  Currently you must hit INFO first if you wish to see, otherwise you are operating in the blind.  Checking on video playback, skipping ahead or going FFWD does pop the OSD and progress bar, so I think we should do this here as well.

PR #832, opened 2023-12-30
DB Schema Update:  Change lambda capture from 'by value' to 'by reference'.

PatrickJMaloney:
This fix corrects a program abort when updating from db version 1302 to 1303.  The problem is that the 'ORDER BY' column name is not specified, causing a fatal SQL error.  The column name should be either 'starttime' or '-starttime' depending on your time zone.

This occurs because the lambda capture is 'by value' and takes on the value at the time it was created (which is empty).  Further updates to the local variable are not available to the lambda expression.

Switching the lambda capture to 'by reference' resolves this issue.

Please apply this change to both the master and fixes/33 branches.

##### Checklist

### DIFF
--- a/mythtv/libs/libmythtv/dbcheck.cpp
+++ b/mythtv/libs/libmythtv/dbcheck.cpp
@@ -474,13 +474,13 @@ static bool doUpgradeTVDatabaseSchema(void)
 
     auto ss2ba = [](const auto & ss){ return QByteArray::fromStdString(ss); };
     auto qs2ba = [](const auto & item){ return item.constData(); };
-    auto add_start_end = [order](const auto & field){
+    auto add_start_end = [&order](const auto & field){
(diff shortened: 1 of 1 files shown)


jhoyt4:
@billmeek - Paul [reported this on the forums](https://forum.mythtv.org/viewtopic.php?p=27026#p27026) and made up the PR to get the fix out to the community.

Please take a look when you get a chance.  (I'm working on some of the other issues in the compile script Paul kindly reported)

kmdewaal:
Looks OK to me, a good catch! This bug is introduced in commit 314498a on Oct 7, 2020.
Will commit this fix to master and v33.

kmdewaal:
Committed in master 7ec562f39820b92c672d430058d2c81cf4844bbd and in v33 6b442547f2746b01f51a051438b4d02db9441b36.

PR #834, opened 2024-01-03
Query audio CD metadata from MusicBrainz

hambre:
This PR adds support to query audio CD metadata from MusicBrainz to display it during playing of an audio CD from within MythMusic.

### DIFF
--- a/mythplugins/configure
+++ b/mythplugins/configure
@@ -302,6 +302,7 @@ exif
 newexif
 dcraw
 cdio
+musicbrainz
 "
 
 DEPEND_LIST="
@@ -356,6 +357,7 @@ MythGame related options:
 MythMusic related options:
   --enable-mythmusic       build the mythmusic plugin [$music]
   --enable-cdio            enable cd playback [$cdio]
+  --enable-musicbrainz     enable fetching cd metadata from musicbrainz [$musicbrainz]
 
 MythNetvision related options:
   --enable-mythnetvision   build the mythnetvision plugin [$netvision]
@@ -715,7 +717,21 @@ if enabled music ; then
             enable cdio; # nop
         else
             disable cdio
-	fi
+        fi
+    fi
+    if enabled musicbrainz ; then
+        if ! check_lib discid/discid.h discid_new -ldiscid ; then
+            disable musicbrainz
+            echo "MusicBrainz support requires libdiscid."
+        elif ! check_lib musicbrainz5/mb5_c.h mb5_query_new -lmusicbrainz5 -lm -lstdc++; then
+            disable musicbrainz
+            echo "MusicBrainz support requires libmusicbrainz5."
+        elif ! check_lib coverart/caa_c.h caa_coverart_new -lcoverart -lm -lstdc++; then
+            disable musicbrainz
+            echo "MusicBrainz support requires libcoverart."
+        else
+            enable musicbrainz; # nop
+        fi
     fi
 
     if ! check_lib lame/lame.h lame_init -lmp3lame ; then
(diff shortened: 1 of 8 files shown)


warpme:
I added req. libs, this pull to current master and it works perfectly fine.
Great work!

mythplugins/configure
```diff
@@ -852,6 +868,16 @@ if enabled music ; then
       echo "#define HAVE_CDIO 1" >> ./mythmusic/mythmusic/config.h
       echo "CONFIG += cdio" >> ./mythmusic/mythmusic/config.pro
       echo "        libcdio        support will be included in MythMusic"
+      if enabled musicbrainz ; then
+        echo "#define HAVE_MUSICBRAINZ 1" >> ./mythmusic/mythmusic/config.h
+        echo "CONFIG += link_pkgconfig" >> ./mythmusic/mythmusic/config.pro
+        echo "CONFIG += musicbrainz" >> ./mythmusic/mythmusic/config.pro
+        echo "PKGCONFIG += libdiscid libmusicbrainz5 libcoverart" >> ./mythmusic/mythmusic/config.pro
```
paul-h, line 875: Should the linked libraries not be libmusicbrainz5cc and libcoverartcc for c++?

hambre: These are the pkgconfig names used to get the respective library names and include paths from pkgconfig. As example taken from the musicbrainz library documentation (http://metabrainz.github.io/libmusicbrainz/5.1.0/):
`g++ -o test_app test_app.cpp pkg-config libmusicbrainz5 --cflags --libs`

paul-h: Strange if I try with what you have everything compiles and links but when I try to run it I get missing symbol errors for both libmusicbrainz5 and libcoverart.

Change to  libmusicbrainz5cc and libcoverartcc and everything works fine?
 
According to the description of the Ubuntu packages libmusicbrainz5cc2v5 is the shared library and libmusicbrainz5-2 is the shared library for the C wrapper functions.

pkg-config contains entries for both libmusicbrainz5 and libmusicbrainz5cc and also for libcoverart and libcoverartcc.

hambre: Oh well :-/, I am on ArchLinux and there is only a single entry for both libmusicbrainz and libcoverart.  Any suggestion on how to solve that?

hambre: After investigating a bit, it seems the split of the c and c++ interfaces was done after the last official release. ArchLinux is using the source tarball of the last official release, so there is only one library present.

hambre: Ok I added a switch to support both version combined and split libs. Could you please try if that works on your side?

paul-h: Thanks it now works for me also.

garybuhrmaster:
I am confused about this entire proposal.  There is already support in the MythTV Metadata Music scripts for musicbrainz.  Why is MythMusic not using that directly, or enhancing those scripts if they are not adequate?  I personally dislike duplicate code paths that (likely) do different things.

hambre:
@garybuhrmaster 

This is what Paul H. wrote in an email to the mythtv-users mailing list:
`The copy of the python MB bindings included with Myth are very old now 
and could do with updating which should be trivial to do. I seem to 
remember the problem with integrating the python script into MythMusic 
is it's an external script so the plugin would have to run an external 
process to do the lookup which I thought was rather messy. Somewhere I 
think I have a patch that used the c++ bindings to MB to do the lookup 
so everything could be done in the MythMusic plugin without having to 
call external scripts. Without looking at the code I'm not sure how far 
I got with it but if anyone is interested I could see if I can find the 
stash.`

I took that as a direction on how to add musicbrainz support to MythMusic.

ijc:
A single discid can be associated with multiple releases e.g. https://musicbrainz.org/cdtoc/_qF9K7DD3fCk3keSxQIR8Dd1ds4- for just one example but it's pretty common in my experience (a common on is the US and European release have the exact same tracks and lengths etc but have different catalogue number, barcode etc, presumably the discs were made in different factories in different markets but the master is the same).

Those are all in the same release group though, so maybe doesn't matter much for the use here, the track names etc would be correct whichever you picked.

But from memory I think I have seen cases where the discid is the same for totally unrelated releases -- IIRC it was a single so had a small number or tracks, which I suppose increases the possibility of a hash collision -- since the hash is just a combination of the track sector offsets/counts fewer tracks means a greater chance of them being the same.

I don't have an example of that to hand though.

ijc:
Here is a clash of two different releases: https://musicbrainz.org/cdtoc/66Ny4dUpzEGf34G0HXSOtByVXq8-

paul-h:
The old CDDB support had the same problem with possible discid clashes which as you say should be rare so it's not a big problem. Definitely a lot better than having no metadata at all :) 

Also while the chosen MB release may not be the correct one as long as the track titles and lengths are correct it doesn't really matter. We don't store or use any of the catalogue number, barcode etc,

paul-h:
I've added the required dependencies to the Ubuntu PPA launchpad builder script for master so those builds should now have support for this.

hambre:
> I've added the required dependencies to the Ubuntu PPA launchpad builder script for master so those builds should now have support for this.

Great! Thanks for merging!

PR #824, opened 2023-12-14
Python 3 cleanups

eli-schwartz:
Roughly divided into two changes:
- remove the use of the `future` PyPI package, which is broken on python 3.12 and unneeded in python 3 entirely.
- perform various code cleanups and drop python2-specific code

The first part is critical to allow running on modern versions of python. The second part I did just because I can, so why not.

PR #838, opened 2024-01-08
hardwareprofile: drop ancient python2 compatibility libraries: "future" and "simplejson"

eli-schwartz:
Two commits:
- hardwareprofile: remove external dependency on simplejson

This uses the requests module and converts requests responses to json using requests' own `.json()` method on responses. For incomprehensible reasons, requests has spent about a decade using either simplejson or the standard library's json module more or less at will, and returning either one or the other exception types. They don't know why they use simplejson, we don't know why they use simplejson. In requests 3 (which will be released in the Year Of The Linux Desktop or when pigs fly, whichever one comes later) simplejson is dropped entirely.

There are innumerable issues discussing the problem on the requests bugtracker, with the general consensus being that it's better to randomly return either one of two different libraries and two different library return types in errors -- because it was historically done that way and people might be depending on it. ??????

Bugs:
https://github.com/psf/requests/pull/710
https://github.com/psf/requests/pull/2516
https://github.com/psf/requests/issues/3052
https://github.com/psf/requests/issues/4169
https://github.com/psf/requests/issues/4842
https://github.com/psf/requests/issues/5794
https://github.com/psf/requests/issues/6084

The awkward workaround is to guarantee that requests' silent behavior of using simplejson *if it is installed* is forcibly triggered by forcibly depending on simplejson, and then catching the simplejson exception.

The better solution here is pretty simple: do not rely on the requests module's automatic json conversion, this is as simple as using the already-imported json module and calling json.loads() on the retrieved content.

Fixes: 1df343e9ab7defa284a73390210a65cf2112f17e
Reimplements: bb154a843b737cc3ad8c1a45fa04a1a3609aff05

- hardwareprofile: remove ancient "future" compatibility library for python2

Ironically, for a package that was intended to provide portability between python2 and python3, it is broken with python 3.12. A better library to use in all cases is "six".

However, mythtv requires python 3.8 for a while now. Using "future.standard_library" is a no-op other than costing a pointless import and being troublesome to actually install.

The hacky copy of six.with_metaclass included in "future" is rewritten to use the pure python3 form of a metaclass.

Part of #824

### DIFF
--- a/.github/workflows/buildmaster.yml
+++ b/.github/workflows/buildmaster.yml
@@ -48,7 +48,7 @@ jobs:
         sudo apt install ccache qt5-qmake qtscript5-dev nasm libsystemd-dev libfreetype6-dev libmp3lame-dev libx264-dev libx265-dev libxrandr-dev libxml2-dev
         sudo apt install libavahi-compat-libdnssd-dev libasound2-dev liblzo2-dev libhdhomerun-dev libsamplerate0-dev libva-dev libdrm-dev libvdpau-dev
         sudo apt install libass-dev libpulse-dev libcec-dev libssl-dev libtag1-dev libbluray-dev libbluray-bdj libgnutls28-dev libqt5webkit5-dev
-        sudo apt install libvpx-dev python3-mysqldb python3-lxml python3-simplejson python3-future python3-setuptools libdbi-perl libdbd-mysql-perl libnet-upnp-perl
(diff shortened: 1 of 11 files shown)


hambre:
Can this possibly be merged to fixes/34 too?

rcrdnalor:
On Sat, Jan 4, 2025 at 10:04 PM hambre ***@***.***> wrote:

> Can this possibly be merged to fixes/34 too?
>

It is in fixes/34, no need to merge.

PR #842, opened 2024-01-10
Reorder structure changes to allow older MythTV databases to migrate cleanly.

PatrickJMaloney:
After my last patch, .25 databases would migrate to .33 without aborting during the process.  However, there were a number SQL errors that still appeared during migration.  The major thing I noticed is that the "Default" recording template was not getting created.  That insert fails due to missing columns and columns that do not allow null values.

The code to add these missing columns and modify nullable attributes is already present, but it is executed after the changes are actually required.  Presumably, some later code changes broke some of the earlier migrations.

I added no new code.  I only moved various SQL structure changes to earlier version updates.  My .25 database now migrates cleanly and the Default recording rule is present after the migration.

Please update master and fixes/33 with these changes.  Thanks!

##### Checklist

### DIFF
--- a/mythtv/libs/libmythtv/dbcheck.cpp
+++ b/mythtv/libs/libmythtv/dbcheck.cpp
@@ -1959,7 +1959,50 @@ static bool doUpgradeTVDatabaseSchema(void)
 "           'smpte2022-1','smpte2022-2'),"
 "  bitrate INT(10) UNSIGNED NOT NULL,"
 "  PRIMARY KEY (iptvid)"
-") ENGINE=MyISAM DEFAULT CHARSET=utf8;"
+") ENGINE=MyISAM DEFAULT CHARSET=utf8;",
+
+            "CREATE TABLE IF NOT EXISTS recgroups ("
+                "recgroupid  SMALLINT(4) NOT NULL AUTO_INCREMENT, "
+                "recgroup    VARCHAR(64) NOT NULL DEFAULT '', "
+                "displayname VARCHAR(64) NOT NULL DEFAULT '', "
+                "password    VARCHAR(40) NOT NULL DEFAULT '', "
+                "special     TINYINT(1) NOT NULL DEFAULT '0',"
+                "PRIMARY KEY (recgroupid), "
+                "UNIQUE KEY recgroup ( recgroup )"
+                ") ENGINE=MyISAM DEFAULT CHARSET=utf8;",
+            // Create the built-in, 'special', groups
+            "INSERT INTO recgroups ( recgroupid, recgroup, special ) VALUES ( 1, 'Default', '1' );",
+            "INSERT INTO recgroups ( recgroupid, recgroup, special ) VALUES ( 2, 'LiveTV', '1' );",
+            "INSERT INTO recgroups ( recgroupid, recgroup, special ) VALUES ( 3, 'Deleted', '1' );",
+            // Copy in the passwords for the built-in groups
+            "DELETE FROM recgrouppassword WHERE password = '';",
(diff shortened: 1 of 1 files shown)


paul-h:
Couldn't this end up breaking more users databases than it fixes? For example what if someone is already on a schema version after the one you moved the updates to but before the schema where those updates where. They now wont get those updates?

PatrickJMaloney:
Yes, that could be.  The solution to this would be something along the lines of keeping what I put in, but also restoring everything I took out.  Only create recgroups if it does not exist.  Column mods being done again shouldn't be a problem.  Insert "ignore" will ignore duplicate keys for recgroups.

Major issue I see is how to tell if a column exists and not add it again.  Not sure what versions the "IF NOT EXISTS" clause is supported for "ADD COLUMN" and that might be MariaDB only.  Handful of local booleans, at worst.

Let me know if I should proceed with those changes.  I will start looking at them anyway.

paul-h:
I'll leave it to kmdewall to sort out the mess :) Think I would revert this for now until a better fix is found. Something like you suggest sounds like it should be OK but I'm no DB expert. Still trying to work out how this problem came about in the first place.

kmdewaal:
I confess to being a bit too quick on this and will revert it. In order to have a better look at it, I would like to have a copy of the database on which the initial problem, i.e. the upgrade failure, can be reproduced.

PatrickJMaloney:
No problem.  I'm reworking this as I suggested.  I can provide a dump of my .25 database.  It's 50MB compressed.  Is that size something that I can attach here?

kmdewaal:
50MB is definitely too much for github. Posting a link to somewhere it can be downloaded from would be great. Some people use pastebin, I use wetransfer.com sometimes.

PatrickJMaloney:
The attached file is a new .25 database with a few recordings.  It will show you all the error types I get on the migration of my full database.  This is a dump from MySQL 5.5.  It will load into 8.x no problem.

[backup25.sql.zip](https://github.com/MythTV/mythtv/files/13897143/backup25.sql.zip)

I have redone the changes.  I kept what I added, but did not remove any of the existing SQL.  I just wrapped the 'add column' statements with a check to make sure they do not exist.  Can I open a new pull request for that?

Thanks, Patrick

kmdewaal:
Thanks for the database backup. I will have a look at it.

About a new fix, yes a new pull request is OK.

A few thoughts about this issue.
The database upgrade code has all been tested very well at the time it was written And everybody ever doing an upgrade from one MythTV version to another has automatically updated his database with the upgrade code.
So it is rather unlikely that there is very much wrong with it.
There are two ways in which I think things can go wrong.

First is rewriting code without testing. 
This was actually the case with the bug that you found in the lambda functions (thanks again!); this was code that was correct when written, but later it was rewritten into modern C++ probably to get rid of compiler/clang-tidy diagnostics. And these changes had not been tested obviously.

The second way that things can go wrong is that the latest SQL do have a slightly different syntax than how it was at the time the database upgrade code was written. And there are slight divergences beteen mysql and mariadb on Linux.
It could also be that sql behaves slightly different on MacOS.

Given these considerations, there should be only minimal changes needed to get the database upgrade working again.

paul-h:
Ah good catch! It was puzzling me how this came about. Makes a lot more sense now :) 

I agree it would be best to try to restore the previous behaviour so it works as it originally did. Do you think we need an additional update to check for a broken DB and to attempt to fix it?

PatrickJMaloney:
PR#844 created.  Just copied SQL this time instead of moving it.  Added checks to stop errors when SQL run a second time.

kmdewaal:
I did have a look at reconstructing the called functions as they were at the time of the initial upgrade, but that is really a huge amount of work and not worth it. More comments, if they arise, in the new PR #844.

PR #844, opened 2024-01-12
Tweak structure changes to allow older MythTV databases to migrate cleanly.

PatrickJMaloney:
Take two on this.  I added a method to DBUtil to check for a column's presence.

.25 databases would migrate to .33 without aborting during the process. However, there were a number SQL errors that still appeared during migration. The major thing I noticed is that the "Default" recording template was not getting created. That insert fails due to missing columns and columns that do not allow null values.

The code to add these missing columns and modify nullable attributes is already present, but it is executed after the changes are actually required. Presumably, some later code changes broke some of the earlier migrations.

This time I removed none of the existing SQL changes. I only copied various SQL structure changes/data updates to earlier db version updates.  I then modified the later ones to not cause an error should they get run a second time.

My .25 database now migrates cleanly and the Default recording rule is present after the migration.

Please update fixes/33 (and master as you see fit) with these changes. Thanks!

##### Checklist

### DIFF
--- a/mythtv/libs/libmythtv/dbcheck.cpp
+++ b/mythtv/libs/libmythtv/dbcheck.cpp
@@ -1971,7 +1971,10 @@ static bool doUpgradeTVDatabaseSchema(void)
     {
         LOG(VB_GENERAL, LOG_CRIT, "Upgrading to MythTV schema version 1302");
 
-        // Adding the default template requires these things.
+        // Adding the default template now requires database table recgroups which was
+        // initially introduced in the upgrade to schema version 1320.
+        // The initial database created with a "clean install" has schema version 1307
+        // so upgrading to schema version 1302 is not often done.
         DBUpdates updates {
             "CREATE TABLE IF NOT EXISTS recgroups ("
                 "recgroupid  SMALLINT(4) NOT NULL AUTO_INCREMENT, "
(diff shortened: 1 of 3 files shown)


mythtv/libs/libmythtv/dbcheck.cpp
```diff
@@ -2687,7 +2748,9 @@ static bool doUpgradeTVDatabaseSchema(void)
             "ADD COLUMN video_avg_bitrate MEDIUMINT UNSIGNED NOT NULL DEFAULT 0, " // Kbps
             "ADD COLUMN video_max_bitrate MEDIUMINT UNSIGNED NOT NULL DEFAULT 0, " // Kbps
             "ADD COLUMN audio_avg_bitrate MEDIUMINT UNSIGNED NOT NULL DEFAULT 0, " // Kbps
-            "ADD COLUMN audio_max_bitrate MEDIUMINT UNSIGNED NOT NULL DEFAULT 0 ;" // Kbps
+            "ADD COLUMN audio_max_bitrate MEDIUMINT UNSIGNED NOT NULL DEFAULT 0 ;", // Kbps
+
+            "ALTER TABLE recorded ADD COLUMN lastplay TINYINT UNSIGNED DEFAULT 0 AFTER bookmark;",
```
kmdewaal, line 2753: Adding field lastplay should be done in the next section, the upgrade to schema 1331 because that is where the field is needed. If somebody would do an upgrade from schema version 1330 then that would still fail.....

mythtv/libs/libmythtv/dbcheck.cpp
```diff
@@ -1970,6 +1970,63 @@ static bool doUpgradeTVDatabaseSchema(void)
     if (dbver == "1301")
     {
         LOG(VB_GENERAL, LOG_CRIT, "Upgrading to MythTV schema version 1302");
+
+        // Adding the default template requires these things.
```
kmdewaal, line 1974: Please add comment like this:
        // Adding the default template now requires database table recgroups which was
        // initially introduced in the upgrade to schema version 1320. 
        // The initial database created with a 'clean install" has schema version 1307
        // so upgrading to schema version 1302 is not often done.

kmdewaal:
Also the git has comments about whitespace:
mythtv/844.patch:126: trailing whitespace.
            (!DBUtil::CheckTableColumnExists(QString("record"), QString("recgroupid")) ? 
mythtv/844.patch:128: trailing whitespace.
            (!DBUtil::CheckTableColumnExists(QString("recorded"), QString("recgroupid")) ? 
mythtv/844.patch:132: trailing whitespace.
            (!DBUtil::CheckTableColumnExists(QString("record"), QString("autoextend")) ? 
mythtv/844.patch:191: trailing whitespace.
            (!DBUtil::CheckTableColumnExists(QString("record"), QString("recgroupid")) ? 
mythtv/844.patch:193: trailing whitespace.
            (!DBUtil::CheckTableColumnExists(QString("recorded"), QString("recgroupid")) ? 
warning: 5 lines add whitespace errors.

mythtv/libs/libmythtv/dbcheck.cpp
PatrickJMaloney: Done.

mythtv/libs/libmythtv/dbcheck.cpp
PatrickJMaloney: Done.

PatrickJMaloney:
Removed whitespace as requested.

kmdewaal:
Thanks for your fixes.

PR #862, opened 2024-02-13
Github workflow: Fix python link error on macOS

jhoyt4:
Homebrew recently shifted to python3.12 as the default python on some
 but not all versions of macOS.  This fix resolves a package linking
 issue experienced when installing python based packages.

##### Checklist

### DIFF
--- a/.github/workflows/buildmaster.yml
+++ b/.github/workflows/buildmaster.yml
@@ -30,6 +30,12 @@ jobs:
             arch: 'arm64'
           - os: 'macos-13'
             arch: 'arm64'
+          - os: 'macos-12'
+            cc: 'gcc'
+          - os: 'macos-13'
+            cc: 'gcc'
+          - os: 'macos-14'
+            cc: 'gcc'
       fail-fast: false
     runs-on: ${{ matrix.os }}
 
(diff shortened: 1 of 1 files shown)


jhoyt4:
marking as a draft until I get the workflow to complete successfully.

garybuhrmaster:
Can we consider moving to one source of truth, and use the ansible repo to install all the needed libraries rather then a different approach for CI here?

I am willing to work with you to (try) to accomplish that if it makes sense?

jhoyt4:
I'm not opposed to going down that path.  On the unofficial macOS CI chain I already do something similar for both [macports](https://github.com/MythTVforMacOS/compileMythTV/blob/main/.github/actions/installPorts/action.yml) and [homebrew](https://github.com/MythTVforMacOS/compileMythTV/blob/main/.github/actions/installBrewPackages/action.yml)

I will note the following comment was added to the workflow code:
      # N.B. These dependencies are for the master branch. Unlike the ansible
      # playlists they do not include old dependencies that may be required for
      # older versions. The list is intended to provide as much code coverage as
      # possible (i.e. enable as many options as possible)

This leads me to believe that not using ansible was intentional

jhoyt4:
I'll also note the issue I'm struggling with currently won't be fixes by ansible.

The issue is an underlying homebrew python issue.  Some of the homebew packages have upgraded to python3.12 but not all.  This leads to a "brew link" issue where both python3.12 and python3.11 try to generate symlinks to ${HB_PREFIX}/bin/python3 etc.

Another issue is that homebrew's runners have a broken python3.12 installed that requires updating.  This version does not clean up all of the symlinks generated when installed (e.g. 2to3-3.12). This breaks the upgrade with a similar link issue.

In reality, we could just wait out homebrew and github fixing these issues upstream.

hambre:
Good to see for that a fix is being worked on, as my PR is having this problem.

jhoyt4:
Finally got through the link issues.  Hopefully I can submit another PR thats backs some of this out in a number of weeks.

jhoyt4:
> Can we consider moving to one source of truth, and use the ansible repo to install all the needed libraries rather then a different approach for CI here?

@garybuhrmaster - I'm still happy to work with you on an ansible based approach in another PR (assuming that's a direction the mythtv devs would like us to go).

jhoyt4:
Three successful CI checks in a row.  I hope this is finally ready.

PR #883, opened 2024-03-17
Fix detection of libglslang in configure

kmdewaal:
Detection of libglslang by the configure script fails on Fedora 39 for MythTV version 34. Detection in master works good, so this is the backport of that code into the configure of MythTV version 34.


kmdewaal:
The intention is that if there are no regressions caused by this in v34 then this fix will also be backported to to v33.

PR #906, opened 2024-06-01
Avoid removing PID filters when using virtual channel tuning

### LINKED ISSUE #905: UpdateFilters called in wrong tune mode for HDHomerun Prime
* **Platform**: All

* **MythTV version**: v35-Pre

* **Package version**:

* **Component**: mythbackend

Using an HDHomerun Prime with a cablecard installed, every time a channel is tuned in, an Error message is written into the mythbackend.log. 

100% of the time. You need a tuner with a cablecard.

Successful tuning in of a channel should not cause generation of an error message. Also, such an error message should be less vague. What tune mode was considered wrong?

### What do you see instead?

Here's an example:

mythbackend: mythbackend[728481]: E HDHRStreamHandler hdhrstreamhandler.cpp:242 (UpdateFilters) HDHRSH\[3\](13122384): **UpdateFilters called in wrong tune mode**

SteveErl:
The first step is to improve the error messages. In two
places the "UpdateFilters called in wrong tune mode"
message has been updated to specify the value of the
wrong tune mode.

After these updates, an example of the error message is:

_mythbackend: mythbackend[779282]: E HDHRStreamHandler
hdhrstreamhandler.cpp:242 (UpdateFilters) HDHRSH\[4\]
(13249773): UpdateFilters called in wrong tune mode, 4_

The tune mode was **hdhrTuneModeVChannel** (= 4), and
we know that **UpdateFilters** was called, but in
hdhrstreamhandler.cpp, the only call to it is avoided for
Vchan tuning.
```
108 void HDHRStreamHandler::run(void)
...
140             UpdateFiltersFromStreamData();
141             if (m_tuneMode != hdhrTuneModeVChannel)
142                 UpdateFilters();
```
Code examination reveals that the offending invocation of
**UpdateFilters()** is in **StreamHandler::RemoveAllPIDFilters()**.

This routine is called unconditionally in
```
108 void HDHRStreamHandler::run(void)
...
183     LOG(VB_RECORD, LOG_INFO, LOC + "RunTS(): " + "shutdown");
184
185     RemoveAllPIDFilters();
```
The solution is to condition the call to **RemoveAllPIDFilters()** in the same way the call to **UpdateFilters()** was previously conditioned. If we're using Virtual Channel Tuning, then don't bother removing filters which were never created in the first place.

After applying this solution, the **_UpdateFilters called in wrong tune
mode_** error messages no longer appear in mythbackend.log.

Resolves #905

##### Checklist

### DIFF
--- a/mythtv/libs/libmythtv/recorders/hdhrstreamhandler.cpp
+++ b/mythtv/libs/libmythtv/recorders/hdhrstreamhandler.cpp
@@ -182,7 +182,10 @@ void HDHRStreamHandler::run(void)
     }
     LOG(VB_RECORD, LOG_INFO, LOC + "RunTS(): " + "shutdown");
 
-    RemoveAllPIDFilters();
+    if (m_tuneMode != hdhrTuneModeVChannel)
+    {
+        RemoveAllPIDFilters();
+    }
 
     {
         QMutexLocker locker(&m_hdhrLock);
@@ -237,7 +240,8 @@ bool HDHRStreamHandler::UpdateFilters(void)
     if (m_tuneMode != hdhrTuneModeFrequencyPid)
     {
         LOG(VB_GENERAL, LOG_ERR, LOC +
-            "UpdateFilters called in wrong tune mode");
+            QString("UpdateFilters called in wrong tune mode, %1")
+                .arg(m_tuneMode));
         return false;
     }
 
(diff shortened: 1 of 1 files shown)


kmdewaal:
I cannot test this myself but it looks good.

PR #935, opened 2024-07-29
Remove NuppelVideoRecorder

ulmus-scott:
See https://github.com/MythTV/mythtv/projects/4#card-46810801

This reduces external code included in MythTV.  MythTV no longer directly uses liblzo or libmp3lame.

@linuxdude42 this is confusing syntax with the unused parameter https://github.com/MythTV/mythtv/blob/b8de924d9a2706fc6f3f1b65707f6cd2ea6a2483/mythtv/libs/libmythtv/CMakeLists.txt#L245 

I have removed the instructions for CMake to build liblzo or libmp3lame, but FFmpeg can use them if they exist, so maybe we want to still build them?

##### Checklist


garybuhrmaster:
Does this mean that those that have (potentially) decades worth of nupplerecorder'ed recordings will suddenly be unable to play those recordings and the project will not, as part of an upgrade, automatically transcode (with 1 to 1 guaranteed quality accuracy) those recordings to something fully supported?

ulmus-scott:
This does not change how  .nuv files are played: they are still demuxed and decoded by FFmpeg.  This removes the ability to create new .nuv files (since FFmpeg only has a demuxer).

Existing .nuv files will still play as is and do not need to be transcoded.  (I only have a single small .nuv sample, but it plays fine).

FFmpeg only uses liblzo for testing.  FFmpeg needs libmp3lame for encoding MP3 audio, but it does have other audio encoders and can copy streams from the input to the output without decoding and encoding them.

garybuhrmaster:
> This does not change how .nuv files are played: they are still demuxed and decoded by FFmpeg. This removes the ability to create new .nuv files (since FFmpeg only has a demuxer).

In the past, removing features (not forced by circumstances outside of the projects control) were announced in the current master (warnings with deprecation when used) and removed in the master next, giving time for people to convert to something different (with documentation as to the recommended alternative).  So, announce now, with STRONG deprecation warnings in the logs in v35 (which will be the first time most will see the warnings), and remove no earlier than v36.

Of course the developers may, again, say, damn the torpedoes, and move faster if they choose.

ulmus-scott:
> In the past, removing features (not forced by circumstances outside of the projects control) were announced in the current master (warnings with deprecation when used) and removed in the master next, giving time for people to convert to something different (with documentation as to the recommended alternative). So, announce now, with STRONG deprecation warnings in the logs in v35 (which will be the first time most will see the warnings), and remove no earlier than v36.
> 
> Of course the developers may, again, say, damn the torpedoes, and move faster if they choose.

I suppose that makes sense, but we need input from the others.

Adding something like:
```
LOG(VB_GENERAL, LOG_ALERT, "NuppelVideoRecorder is deprecated and will be "
"removed in a future version of MythTV.  Adjust transcode settings to use "
"FFmpeg or change recorder settings.");
```
to the NuppelVideoRecorder constructor could work, but I have no idea what the changed settings would be.

bennettpeter:
I do not believe the nuppelvideo recorder works any more. A couple of years ago I configured a USB camera in mythtv as a capture card to see whether  it would work. When I created a test recording, the log said it was using nuppelvideo, but nothing was created and the recording failed with numerous errors.

linuxdude42:
A mailing list comment from 2017 says they thought there was agreement to drop support for non-hardware recorders after 29, so this removal has been discussed off and on for quite some time.  I know I brought it up in 2019.

ulmus-scott:
If it doesn't work then there is no problem removing it.  @bennettpeter can you test it again?

Could MpegRecorder or V4L2encRecorder be used instead?  (I couldn't find any documentation on what those were for.)

If we did want to support a software encoding recorder, it should use FFmpeg.  If we don't, it should be possible to set up an external recorder (probably using FFmpeg) to do that.

bennettpeter:
An external recorder can be used. i tested it and put it here : https://github.com/bennettpeter/MythTV-V4L2Capture
I did not implement the analog capture cards in the new web setup. They are grayed out. The only way to set them up would be with mythtv-setup. When I created the web setup I tested mythtv-setup and decided that it was not worth while adding them to the web setup.

ulmus-scott:
@bennettpeter did you have a chance to test if NuppelVideoRecorder works?

ulmus-scott:
Rebased onto master, fixing up the QMake project files from moving ffmpeg-mmx.h and removing the now unused `libmythtv/recorders/vbitext` directory.

@bennettpeter did you have a chance to test if NuppelVideoRecorder works?

ulmus-scott:
Has anyone been able to test if NuppelVideoRecorder works in master?

linuxdude42:
The latest changes compile when using "make", and when using "cmake" for linux builds.  They fail for all "cmake" android builds because of two remaining references to lzo, [here](https://github.com/MythTV/mythtv/blob/0970091f4d93e2b97b8b774a07ae8fa36c4f1c21/mythtv/cmake/MythFindPackages.cmake#L204) and [here](https://github.com/MythTV/mythtv/blob/0970091f4d93e2b97b8b774a07ae8fa36c4f1c21/mythtv/libs/libmythtv/cmake/Backend.cmake#L11).

Cmake still needs to build the lame library for the mythmusic plugin. The compiles will complete without it, but the plugin will be disabled.

linuxdude42:
I'm building it for android.

Issue #978, opened 2024-11-29
Forced HDMV PGS subtitles stay on screen past their intended time slot


Glidos:
* **Platform**: Linux

* **MythTV version**: 34 and master

* **Package version**: N/A

* **Component**: mythfrontend

### What steps will reproduce the bug?
Play a video that has HDMV PGS subtitles with at least one forced track. Don't enabled subtitles.

### How often does it reproduce? Is there a required condition?
Always. No condition.

### What is the expected behaviour?
Each subtitle should stay on screen for a shortish period of time (unless it's an unusual file).

### What do you see instead?
 At the end of a sequence of forced subitles, the last of the sequence will stay on screen.

### Additional information
I have a fix for this, which I will post soon.

PR #979, opened 2024-11-29
Fix failure to clear last forced subtitle in a sequence

Glidos:
Some types of subtitle, rather than being tagged with a duration, rely on the next in sequence to clear them, specifically with an empty subtitle at the end of the sequence. AddAVSubtitle has to handle forced subtitles, causing them to be displayed even when subtitles are off. It was detecting a subtitle being forced by looking for a "forced" flag amongst the subtitle's rectangles. That failed to detect the case of an empty subtitle. Hence the last non-empty forced subtitle was not being cleared by the subsequent empty one. This commit changes the detection condition, using instead a flag that denotes whether the subtitle is from the currently selected forced track.

This fixes Issue #978

##### Checklist


kmdewaal:
I did not reproduce this issue but the fix looks OK. Will be committed now, if anybody complains it will be reverted.

PR #984, opened 2024-12-04
fix QT>=6.4 segfault by always setting surface type

### LINKED ISSUE #754: Qt6: Current master segfaults on qt-xcb(x11) and qt-wayland qpa on Qt6.4+
* **Platform**:
MiniMyth2; qt6 build from sources

* **MythTV version**:
current master (tested on v34-Pre-d54bb609dac)

* **Package version**:

* **Component**:
Qt6

Build myth-master on system with Qt6 newer than 6.3 and run mythtv with drawing to x11 or wayland.
Drawing to eglfs qpa works ok.
As issue:
1. starts to happen only when Qt6 is newer than 6.3
2. Verification of relevant qt6 dependencies looks ok (libs vers. verified with documentation; no any warnings from Qt configure)
3. all version newer than 6.3 (6.4.0 from 7 moths ago) are failing - I don't believe it is 7+ months long-standing Qt6 bug.
I suspect issue might be rather related to i.e. https://doc.qt.io/qt-6/opengl-changes-qt6.html starting to manifetst incompatibility between mythtv OpenGL code and Qt6 current OpenGL support?  

always

correct operation on Qt6.4 and Qt6.5

### What do you see instead?
Qt 6.5.0 with drawing to X11 gives me segfault like this: https://gist.github.com/warpme/95abdde87d468ce8cf4bb2416159e6d4

The same binaries (qt6, mythtv) are working ok with drawing to eglfs.
(https://gist.github.com/warpme/cdc7e9a23bca4cb10c0433d880cc288d).

### Additional information
I tested following Qt versions:
(each time only qt was rebuild)

- 5.15.9 - ok
- 6.2.3 - ok
- 6.3.2 - ok (last in v6.3 line)
- 6.4.0 - nok (first in v6.4 line)
- 6.4.3 - nok (last in v6.4 line)
- 6.5.0 - nok

mitchcapper:
See #754 for details but this is the recommendation from QT dev @alpqr  that 100% WFM.   Note, as I mention there officially the old behavior would be best replicated by setting this to the internal value `QSurface::RasterGLSurface` from https://doc.qt.io/qt-6/qsurface.html#SurfaceType-enum  as this is what QT 6.3 and lower did.  If you want me to update the PR to that I can.  I don't have osx to test there.  

Clearly people are running mythtv with QT > 6.3 so I am not sure why only some users hit this bug.   Maybe it is only if they have hardware rendering enabled? Given how long it has stood though there must be something scoping it to just a select few of us.

##### Checklist

- [ x] documentation added/updated/removed where necessary

Fixes #754

### DIFF
--- a/mythtv/libs/libmythui/opengl/mythrenderopengl.cpp
+++ b/mythtv/libs/libmythui/opengl/mythrenderopengl.cpp
@@ -541,12 +541,10 @@ void MythRenderOpenGL::SetWidget(QWidget *Widget)
         return;
     }
 
-#ifdef Q_OS_ANDROID
     // Ensure surface type is always OpenGL
     m_window->setSurfaceType(QWindow::OpenGLSurface);
     if (native && native->windowHandle())
         native->windowHandle()->setSurfaceType(QWindow::OpenGLSurface);
-#endif
 
 #ifdef USING_QTWEBENGINE
     auto * globalcontext = QOpenGLContext::globalShareContext();


jhoyt4:
This is confirmed to work on macOS with qt 6.8

linuxdude42:
Doesn't work for me on Fedora.

mitchcapper:
@linuxdude42 please post full startup logs and run mythfrontend with env var `QT_DEBUG_PLUGINS=1`

linuxdude42:
I don't get any additional output when running the qt6 based image with QT_DEBUG_PLUGINS set.  I do get output for the qt5 image, so I know I've set the variable correctly.  Does the attached excerpt from a strace output help? It shows that the files for the egl plugins are being opened.
[log.txt](https://github.com/user-attachments/files/18216036/log.txt)

mitchcapper:
sorry try this
```export QT_DEBUG_PLUGINS=1
mythfrontend -v general,gpu,gui &> myth.log```   then attach the myth.log

**Comment by linuxdude42** [comment]
I hope this helps, but it still doesn't have any qt debug output.
[myth.log](https://github.com/user-attachments/files/18217025/myth.log)

**Comment by mitchcapper** [comment]
Yeah that is pretty lacking, just to confirm before applying this PR you were getting "Qt: QOpenGLContext::makeCurrent() called with non-opengl surface " ?

the 300D error is the same that @warpme gets currently.   The EGL error is actually "EGL_BAD_SURFACE" which would imply certainly a similar error to what Qt itself was saying without.   

Can you try your graphics cards proprietary driver?   Does the display server log have anything potentially useful in it?

**Comment by linuxdude42** [comment]
Yes, before the patch I get:

    Qt: QOpenGLContext::makeCurrent() called with non-opengl surface 0x1753c9e0

and after the patch I get:

    Qt: QEGLPlatformContext: eglMakeCurrent failed: 300d

My laptop has Intel's UHD Graphics 630 I am using the Intel i915 driver.  I don't see anything useful in the Xorg.0.log file.

I'm happy to compile in any debugging code you think would be useful.  I would very much like to get this resolved so we can (hopefully) transition the default compile from qt5 to qt6 after the next release. Feel free to contact me directly.

**Comment by linuxdude42** [comment]
I see the same 300d error on one of my production frontends.  That system is an Intel NUC6 with HD Graphics 500 video using the i915 driver.

**Comment by mitchcapper** [comment]
In reviewing what was said in #754  at least @kmdewaal also got this error with intel graphics and @warpme is thought to have as well.

@jhoyt4 can you confirm what graphics card is driving the box it worked on?   I assume plenty of people are using these on mini pcs thus the intel only graphics but if it does work with amd/nvidia proprietary drivers that would be interesting.  At a minimum may make narrowing the problem easier.

**Comment by jhoyt4** [comment]
Apple GPU on and M1 and M3max

**Comment by linuxdude42** [comment]
Another data point.  Works on my old 2014 Mac Mini running macOS 12.7.6 and using Intel Iris graphics.

**Comment by kmdewaal** [comment]
Another data point: boty menu and  playback is OK under the following conditions:

- export MYTHTV_NO_EGL=1
 (see mythdisplay.cpp:1222)
- use FFmpeg decode ("Normal") instead of VAAPI hardware acceleration

This is with Fedora 41, Qt 6.8.0, master with this fix included.

**Comment by warpme** [comment]
> * export MYTHTV_NO_EGL=1

Indeed - this confirms (of me): issue is within egl based x11 integration in qt6(xcb_egl).
I think - as of my last report in https://bugreports.qt.io/browse/QTBUG-113728?focusedId=848677&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-848677 - issue is within qt6 - not in mythtv. 

As x11 with xcb_glx and eglfs are working, I'm concluding:
1. mythtv correclty operates with qt6 gl subsystem (eglfs; xcb_glx are working)
2. issue is in internal qt6 xcb_egl gl integration layer

Before Qt6.4 all was working ok - so for me all this points to regression in qt6+   

It is a bit sad for me that Qt team seems to be not interested in resolving this issue...

**Comment by kmdewaal** [comment]
Applying this fix and then testing it with Qt5 gives the following new error message in the mythfrontend log:

`2025-01-18 20:09:28.765530 I [843752/843752] CoreContext mythcommandlineparser.cpp:2932:operator()  Qt: Attempted flush to non-raster surface QWidgetWindow(0xbda4df0, name="mainwindowWindow") of type QSurface::OpenGLSurface (consider using Qt::WA_PaintOnScreen to exclude from backingstore sync)
`

Doing what the error message says on the parent widget with the following code in mythrenderopengl.cpp:560:
```
    if (!create())
        LOG(VB_GENERAL, LOG_CRIT, LOC + "Failed to create OpenGLContext!");
    else
    {
        Widget->setAttribute(Qt::WA_PaintOnScreen);
        if (native)                                         // KdW test
            native->setAttribute(Qt::WA_PaintOnScreen);     // KdW test
    }
```

This does fix the error message in Qt5.

I think the patch from this ticket combined with the code snippet above can be applied to master.
I have not seen adverse effects with Qt5 and the patch, as mentioned, does fix the Qt6 problem both for VDPAU based systems and for Piotr's embedded systems.

It does however NOT fix the abort in Qt6 for Intel/VAAPI based systems. This still needs to be done.

Issue #1035, opened 2025-02-10
MythTV crashes trying to play bluray disc


linuxdude42:
Trying to play a bluray disc results in a segfault in the FFmpeg code.  This started happening with the FFmpeg 7 upgrade.  If I check out and build commit b1b19f4489 from the end of December, the problem doesn't happen.

* **Platform**:

Linux 6.12.6-200.fc41.x86_64 #1 SMP PREEMPT_DYNAMIC Thu Dec 19 21:06:34 UTC 2024 x86_64 GNU/Linux

* **MythTV version**:

master: 

* **Component**:

FFmpeg

### What steps will reproduce the bug?

1) Insert a bluray disc.
2) Ensure its mounted where MythTV can see it
3) Go to the Optical Disc menu and choose Play Optical Disc.

### How often does it reproduce? Is there a required condition?

Every time.

### What is the expected behaviour?

Bluray plays.

### What do you see instead?

Crash in FFmpeg.

### Additional information

Here's a stack trace of the crash:

Thread 103 "Decoder" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffeb3fff6c0 (LWP 2746054)]
0x00007ffff184e4a6 in has_codec_parameters (st=st@entry=0x2169d80, errmsg_ptr=0x0) at libavformat/demux.c:2071
2071	        if (sti->info->found_decoder >= 0 && avctx->pix_fmt == AV_PIX_FMT_NONE)
(gdb) bt
#0  0x00007ffff184e4a6 in has_codec_parameters (st=st@entry=0x2169d80, errmsg_ptr=0x0) at libavformat/demux.c:2071
#1  0x00007ffff1854b3a in avformat_find_stream_info (ic=0xaa4d40, options=0x0) at libavformat/demux.c:2601
#2  0x00007ffff7690d06 in AvFormatDecoder::FindStreamInfo (this=this@entry=0x215a540) at decoders/avformatdecoder.cpp:885
#3  0x00007ffff759cdfa in MythBDDecoder::StreamChangeCheck (this=0x215a540) at Bluray/mythbddecoder.cpp:71
#4  0x00007ffff76a8512 in AvFormatDecoder::GetFrame (this=0x215a540, decodetype=kDecodeAV, Retry=@0x7ffeb3ffe54f: false) at decoders/avformatdecoder.cpp:4624
[...]

PR #1038, opened 2025-02-11
Fix Blu-ray crashes

ulmus-scott:
call avformat_find_stream_info() only once

It frees FFStream::info, which will cause a segmentation fault on a second call
since info will be NULL.

silence_ffmpeg_logging is never set and -1 is not an AVERROR code (depends on
the values used by errno.h), so those can be removed.

Closes https://github.com/MythTV/mythtv/issues/1035.

##### Checklist

### DIFF
--- a/mythtv/libs/libmythtv/Bluray/mythbddecoder.cpp
+++ b/mythtv/libs/libmythtv/Bluray/mythbddecoder.cpp
@@ -68,7 +68,6 @@ void MythBDDecoder::StreamChangeCheck(void)
         QMutexLocker locker(&m_trackLock);
         Reset(true, false, false);
         CloseCodecs();
-        FindStreamInfo();
         ScanStreams(false);
         m_streamsChanged=false;
     }
--- a/mythtv/libs/libmythtv/decoders/avformatdecoder.cpp
+++ b/mythtv/libs/libmythtv/decoders/avformatdecoder.cpp
@@ -126,8 +126,6 @@ static constexpr int16_t kMaxVideoQueueSize = 220;
 static constexpr ssize_t kMaxVideoQueueSize = 220;
 #endif
 
-static bool silence_ffmpeg_logging = false;
-
 static QSize get_video_dim(const AVCodecContext &ctx)
 {
     return {ctx.width >> ctx.lowres, ctx.height >> ctx.lowres};
@@ -265,9 +263,6 @@ static bool StreamHasRequiredParameters(AVCodecContext *Context, AVStream *Strea
 
 static void myth_av_log(void *ptr, int level, const char* fmt, va_list vl)
 {
-    if (silence_ffmpeg_logging)
-        return;
-
     if (VERBOSE_LEVEL_NONE())
         return;
 
--- a/mythtv/libs/libmythtv/decoders/avformatdecoder.cpp
+++ b/mythtv/libs/libmythtv/decoders/avformatdecoder.cpp
@@ -879,21 +879,6 @@ extern "C"
     }
 }
 
-int AvFormatDecoder::FindStreamInfo(void)
-{
-    m_avCodecLock.lock();
-    int retval = avformat_find_stream_info(m_ic, nullptr);
-    m_avCodecLock.unlock();
-    silence_ffmpeg_logging = false;
-    // ffmpeg 3.0 is returning -1 code when there is a channel
-    // change or some encoding error just after the start
-    // of the file, but is has found the correct stream info
-    // Set rc to 0 so that playing can continue.
-    if (retval == -1)
-        retval = 0;
-    return retval;
-}
-
 /**
  *  OpenFile opens a ringbuffer for playback.
  *
@@ -1042,7 +1027,9 @@ int AvFormatDecoder::OpenFile(MythMediaBuffer *Buffer, bool novideo,
         m_ic->max_analyze_duration = 60LL * AV_TIME_BASE;
 
         m_avfRingBuffer->SetInInit(m_livetv);
-        err = FindStreamInfo();
+        m_avCodecLock.lock();
+        err = avformat_find_stream_info(m_ic, nullptr);
+        m_avCodecLock.unlock();
         if (err < 0)
         {
             LOG(VB_GENERAL, LOG_ERR, LOC + QString("Could not find codec parameters for '%1'").arg(filename));
--- a/mythtv/libs/libmythtv/decoders/avformatdecoder.h
+++ b/mythtv/libs/libmythtv/decoders/avformatdecoder.h
@@ -120,7 +120,6 @@ class AvFormatDecoder : public DecoderBase
     int SetTrack(uint Type, int TrackNo) override;
 
     int ScanStreams(bool novideo);
-    int FindStreamInfo(void);
 
     int  GetNumChapters() override; // DecoderBase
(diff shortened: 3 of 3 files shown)


linuxdude42:
Well, it doesn't crash, but it doesn't play any video and the audio stutters badly. From the output, it looks like its constantly re-processing what streams are on the disc.  I've attached a [log file](https://github.com/user-attachments/files/18745301/log.txt).

ulmus-scott:
The log is expected since MythBDDecoder::StreamChangeCheck() calls CloseCodecs():

https://github.com/MythTV/mythtv/blob/3aceb1c39fc474a8ef944cea1e9eb8d4b77e39a4/mythtv/libs/libmythtv/Bluray/mythbddecoder.cpp#L70

This is originally from https://github.com/MythTV/mythtv/commit/d1e7906677e6776455791a0cd43e00ca441b5947

That causes new AVCodecContexts to have to be created for each stream, which may be causing the stuttering.

Unfortunately, I can't test this myself since I don't have a computer Blu-ray player.  Could you test reapplying https://github.com/MythTV/mythtv/commit/616074a0a5f4fa9ad7241e673d6693cbd7b5e310 and additionally removing that call with this pull request?

I do want to look at https://github.com/MythTV/mythtv/commit/616074a0a5f4fa9ad7241e673d6693cbd7b5e310 again before it is reapplied, however, since I don't know why it is calling streams_changed() every 4-5 ms for Blu-rays.

However, the code in master is still incorrect since "avformat_find_stream_info() is indeed not supposed to be called more
than once (on the same AVFormatContext)." https://ffmpeg.org/pipermail/ffmpeg-devel/2025-February/339642.html

ulmus-scott:
Could you provide a log with `-v playback:debug,libav:trace` and the following change applied?
```
diff --git a/mythtv/libs/libmythtv/decoders/avformatdecoder.cpp b/mythtv/libs/libmythtv/decoders/avformatdecoder.cpp
index e6126b02f5..91bc8cd827 100644
--- a/mythtv/libs/libmythtv/decoders/avformatdecoder.cpp
+++ b/mythtv/libs/libmythtv/decoders/avformatdecoder.cpp
@@ -854,11 +854,11 @@ void AvFormatDecoder::streams_changed(void *data, int avprogram_id)
 
     int cnt = decoder->m_ic->nb_streams;
 
+    auto* program = decoder->get_current_AVProgram();
     LOG(VB_PLAYBACK, LOG_INFO, LOC +
-        QString("streams_changed 0x%1 -- program_number %2 stream count %3")
-            .arg((uint64_t)data,0,16).arg(QString::number(avprogram_id), QString::number(cnt)));
+        QString("streams_changed 0x%1 -- program %2 program_number %3 stream count %4")
+            .arg((uint64_t)data,0,16).arg(pointerToQString(program), QString::number(avprogram_id), QString::number(cnt)));
 
-    auto* program = decoder->get_current_AVProgram();
     if (program != nullptr && program->id != avprogram_id)
     {
         return;

```

I suspect https://github.com/MythTV/mythtv/blob/3aceb1c39fc474a8ef944cea1e9eb8d4b77e39a4/mythtv/external/FFmpeg/libavformat/mpegts-mythtv.c#L2513 is returning NULL and the check for this that I removed in https://github.com/MythTV/mythtv/commit/b5ae1fcd2d8067e68eb1eca7ecf04b2597dcbaa1#diff-7bfc6a356e2b4469d533e0265810de2b3b7cbd3d5f436da3f6b65db08188f949L2749 is necessary.

linuxdude42:
With b5ae1fc added back, the call to CloseCodecs() removed, and your patch:

I get [this output](https://github.com/user-attachments/files/18774050/output.txt) before the crash in has_codec_parameters().

ulmus-scott:
I had meant for you to test with https://github.com/MythTV/mythtv/commit/b5ae1fcd2d8067e68eb1eca7ecf04b2597dcbaa1, https://github.com/MythTV/mythtv/pull/1038/commits/a757f3b7b18bb3fbfac5bd45b1e631e527acd672, the modified logging, and additionally removing the call to CloseCodecs() in MythBDDecoder::StreamChangeCheck() to see if it was still stuttery.

Regardless, the log you provided demonstrates that while the PMT version_number is a reliable indicator for this use with my broadcast TS samples, including the Finnish YLE samples that triggers it, it is not a reliable indicator for your Blu-ray disk.

It looks like there is a new PMT version for each frame, which seems odd.  Perhaps the Blu-ray is sending data via the descriptors in the PMT for each frame?

Out of curiosity, what Blu-ray are you testing with and does the issue occur with other Blu-ray disks?

---

It had seemed possible to reduce our downstream changes with https://github.com/MythTV/mythtv/commit/b5ae1fcd2d8067e68eb1eca7ecf04b2597dcbaa1, however, it should stay reverted.

The two commits in this pull request are still valid and correct and should be merged.

ulmus-scott:
> I had started testing with [b5ae1fc](https://github.com/MythTV/mythtv/commit/b5ae1fcd2d8067e68eb1eca7ecf04b2597dcbaa1), [a757f3b](https://github.com/MythTV/mythtv/commit/a757f3b7b18bb3fbfac5bd45b1e631e527acd672), the modified logging, and additionally removing the call to CloseCodecs() in MythBDDecoder::StreamChangeCheck(). The results were very interesting.
> 
>     * The original disc, Rio, now plays audio, but the video is a series of stills spaced about 10 seconds apart.
> 

Since _Rio_ repeatedly calls `streams_changed()` with https://github.com/MythTV/mythtv/commit/b5ae1fcd2d8067e68eb1eca7ecf04b2597dcbaa1, I suspect the video being temporally widely spaced frames may be related to the call to `Reset()` in `MythBDDecoder::StreamChangeCheck()`.

> 
> I tried a couple of other discs.
> 
>     * The recent Dune movie plays properly.
> 
>     * The animated movie Onward plays fluidly, but it plays the descriptive audio track. (Same as older code.)
> 
>     * The Expanse, Season 4 Disc 1 plays properly.
> 
> 
> When I came back to do more testing, I saw that you had pushed changes, so I reset my tree and merged your changes. Rio now plays audio and video, but it suffers some blockiness. The other three movies still play the same.

I just rebased onto master because I don't think https://github.com/MythTV/mythtv/commit/b5ae1fcd2d8067e68eb1eca7ecf04b2597dcbaa1 should be reapplied but that those two commits are still valid.

For _Onward_, you would probably have to look at what tracks are available and the heuristics in `AvFormatDecoder::AutoSelectAudioTrack()` and `AvFormatDecoder::selectBestAudioTrack()` if you want it to select a different track by default.

I suspect the blockiness with _Rio_ is unrelated to the changes here and would also be present in master since it should not be calling streams_changed() with https://github.com/MythTV/mythtv/commit/b5ae1fcd2d8067e68eb1eca7ecf04b2597dcbaa1 reverted.

linuxdude42:
Onward has the same audio track selection before/after all the FFmpeg changes, so that's a different issue.

To test Rio again, I rebuilt from my known working point ([b1b19f4](https://github.com/MythTV/mythtv/commit/b1b19f4489d8b3d35a252c0f8a961e58aa202afe)) from December. I do see some blockiness at a later point in the movie, so the blockiness is probably also not a result of the FFmpeg changes.

I'll commit these the two patches here as they do prevent crashing.

Thanks for figuring this out.

Issue #1039, opened 2025-02-14
Mythtv v35 fails decode with Nvidia NVDEC


RyanGoat:
* **Platform**:
Debian 12.9
Linux htpc 6.1.0-31-amd64 # 1 SMP PREEMPT_DYNAMIC Debian 6.1.128-1 (2025-02-07) x86_64 GNU/Linux

* **MythTV version**:
MythTV Version : v35.0+fixes.20250213.ac34c663b2-0ubuntu0
MythTV Branch : fixes/35
Network Protocol : 91
Library API : 35.20220913-1
QT Version : 5.15.8
Options compiled in:
 linux profile use_hidesyms using_alsa using_oss using_pulse using_pulseoutput using_backend using_bindings_perl using_bindings_python using_bindings_php using_dvb using_firewire using_frontend using_hdhomerun using_satip using_vbox using_ceton using_joystick_menu using_libcec using_libcrypto using_gnutls using_libdns_sd using_libxml2 using_lirc using_mheg using_opengl using_egl using_qtwebkit using_qtscript using_qtdbus using_taglib using_v4l2 using_v4l2prime using_x11 using_system_libbluray using_system_libudfread using_systemd_notify using_systemd_journal using_drm using_bindings_perl using_bindings_python using_bindings_php using_freetype2 using_mythtranscode using_opengl using_egl using_drm using_vaapi using_nvdec using_vdpau using_ffmpeg_threads using_mheg using_libass using_libxml2 using_libmp3lame

* **Package version**:
compiled from source

* **Component**:
mythfrontend

### What steps will reproduce the bug?
1. Configure video playback to use NVIDIA NVDEC acceleration
2. Attempt to playback any video
3. After a delay mythfrontend reports, "Video Frame Buffering Failed Too Many Times"

### How often does it reproduce? Is there a required condition?
100% reproducible.  Fails every time.

### What is the expected behaviour?
Videos should play without issue.

### What do you see instead?
Videos fail to play.

### Additional information
Videos played correctly with NVDEC on the same system with MythTV v34.

ulmus-scott:
Please attach a log trying to play a video with `mythfrontend -v playback:trace,libav:trace --logpath=/tmp`.

Also, what Nvidia driver version are you using?

RyanGoat:
Here is the requested log:
[mythfrontend.20250215024209.1066651.log](https://github.com/user-attachments/files/18807573/mythfrontend.20250215024209.1066651.log)

[...]

PR #1040, opened 2025-02-15
rollback ffnvcodec to 12.1.14.1

ulmus-scott:
This reduces the minimum Nvidia driver level to 530 from 550.  Nvidia lists versions 535 and 550 for download for Linux.

Debian currently only packages 535 for Debian 12 and above.

Fixes https://github.com/MythTV/mythtv/issues/1039.  This should also be applied to fixes/35

##### Checklist

### DIFF
--- a/mythtv/external/nv-codec-headers/CMakeLists.txt
+++ b/mythtv/external/nv-codec-headers/CMakeLists.txt
@@ -10,7 +10,7 @@ cmake_minimum_required(VERSION 3.20)
 #
 project(
   nv-codec-headers
-  VERSION 12.2.72.1
+  VERSION 12.1.14.1
   LANGUAGES C)
 
 include(GNUInstallDirs)
--- a/mythtv/external/nv-codec-headers/README
+++ b/mythtv/external/nv-codec-headers/README
@@ -3,5 +3,5 @@ FFmpeg version of headers required to interface with Nvidias codec APIs.
 Corresponds to Video Codec SDK version 12.0.16.
 
 Minimum required driver versions:
-Linux: 550.54.14 or newer
-Windows: 551.76 or newer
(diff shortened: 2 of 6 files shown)


bennettpeter:
I cannot recrete the error that the user had, see #1039. Playback is fine with nvidia 550. Do we really want to revert to older software?

ulmus-scott:
https://www.mythtv.org/wiki/Coding_Standards lists Debian stable, which is currently 12, as supported, and Debian 12 only has a package for the Nvidia 535 driver.

When I updated ffnvcodec for FFmpeg 7.1, I didn't have a specific reason to use 12.2.72.1, just that it was the newest and I thought the 550 driver had been out for long enough that it would have been widely available, which apparently was incorrect.  I only needed a different version since FFmpeg no longer accepted the 9.1.23.2 version and would disable NVDEC/NVENC completely otherwise.

12.2.72.1 lists 550 as the minimum driver version, so I would expect it to work.  12.1.14.1 lists 530 as the minimum driver version.

I don't think there is any significant difference for NVDEC between those versions, but, if 12.1.14.1 works with 535 and 12.2.72.1 doesn't, I think we should roll ffnvcodec back to 12.1.14.1.

ffnvcodec is just a copy of the Nvidia headers that FFmpeg needs to dynamically load the CUDA library for NVDEC and NVENC, so, since the newer drivers still accept the older interface, I don't think it is really "older software", just that it is allowing FFmpeg to work with older drivers.

bennettpeter:
OK I will test with 535 without and with the patch.

ulmus-scott:
I am closing this since #1039 has been resolved and this only effects NVENC, which is disabled by configure but enabled by cmake, pending discussion on the mailing list.

PR #1044, opened 2025-02-16
remove libmythupnp Rtti and merge remnants of libmythservicecontracts into mythfrontend

ulmus-scott:
@bennettpeter Rtti was unused, but the mythfrontend stuff still works (except that using port 6551 prevents mythfrontend from shutting down).

##### Checklist


bennettpeter:
@ulmus-scott I did a quick test of mythfrontend without this pull request and it did shut down successfully after I ran one service (http://localhost:6551/Frontend/GetActionList). Is there a special condition under which it does not shut down, or are you saying that this pull request causes it to not shut down?

ulmus-scott:
@bennettpeter This pull request should not effect anything on port 6551.

To prevent proper shutdown:
1. Launch `mythfrontend` from a terminal.
2. Either:
2.a. Open http://localhost:6551/MythFE/GetScreenShot or http://localhost:6551/MythFE/GetServDesc
2.b. Open http://localhost:6551/MythFE/GetRemote or http://localhost:6551/MythFE/GetActionTest
2.b.1. Sometimes I had to also press one of the buttons, but not when I tested again just now.
3. Return to the mythfrontend window and exit normally.
4. The window will disappear, but the process in the terminal is still running with the last log message being: "Shutting down UPnP client...".  It will stay there until terminated with ^C^C.

Without accessing port 6551, the next log message occurs after about 1 second and "Exiting" about 2 seconds after that.  The terminal then presents the command prompt.

ulmus-scott:
From my post to mythtv-dev:

The use of any of the pages on port 6551 cause mythfrontend to hang indefinitely when closing.

It waits indefinitely somewhere in https://github.com/MythTV/mythtv/blob/0475f039f91ee8726ce3a955f54149665fe201da/mythtv/libs/libmythbase/mthreadpool.cpp#L267 called from
https://github.com/MythTV/mythtv/blob/0475f039f91ee8726ce3a955f54149665fe201da/mythtv/libs/libmythupnp/httpserver.cpp#L176 from
https://github.com/MythTV/mythtv/blob/0475f039f91ee8726ce3a955f54149665fe201da/mythtv/programs/mythfrontend/mediarenderer.cpp#L180 from
https://github.com/MythTV/mythtv/blob/0475f039f91ee8726ce3a955f54149665fe201da/mythtv/programs/mythfrontend/mythfrontend.cpp#L335

bennettpeter:
There was a similar problem of not shutting down because of a thread not terminated in mythbackend a couple of years ago. Somebody, I think maybe Stuart, fixed it.

ulmus-scott:
I fixed the shutdown issue in https://github.com/MythTV/mythtv/pull/1066.  The presentation in master has changed slightly since I removed gContext: the GUI is freezes/the screen doesn't update on alt-tab or changing back from a different virtual desktop since the window is now destroyed after the UPnP threads are shut down.

PR #1066, opened 2025-04-03
Prevent deadlock in MThreadPool

ulmus-scott:
@bennettpeter This resolves the shutdown issue mentioned in https://github.com/MythTV/mythtv/pull/1044

The deadlock can be triggered in mythfrontend by using http://localhost:6551/MythFE/GetRemote and then exiting within 10 seconds.

For a version with additional logging I used to diagnose this, see https://github.com/ulmus-scott/mythtv/tree/upnp_shutdown_log

##### Checklist

### DIFF
--- a/mythtv/libs/libmythbase/mthreadpool.cpp
+++ b/mythtv/libs/libmythbase/mthreadpool.cpp
@@ -103,12 +103,9 @@ using MPoolQueues = QMap<int, MPoolQueue>;
 class MPoolThread : public MThread
 {
   public:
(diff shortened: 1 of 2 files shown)


ulmus-scott:
I think QThreadPool's implementation is better architected in that it has only one mutex in the QThreadPool instead of that one and one per pool thread and that the pool threads request from the QThreadPool a QRunnable to run instead of being given one by the thread pool; however, I don't plan on changing this.

PR #1080, opened 2025-04-30
MythContext::Impl::FindDatabase(): read the database parameters after they have been changed

ulmus-scott:
This fixes bugs from e10ba67e29e5df144dfb5eb0a383c4460bff9c5a where the parameters read from config.xml (or the defaults if it doesn't exist) are always saved back to config.xml.

goto NoDBfound skipped over the initialization of the moved dbParams, so was copied to its two uses.

My testing showed `mythbackend` and `mythtv-setup` could create config.xml.  I do not have a setup for a remote frontend, so I could not get `mythfrontend` to create config.xml.
##### Checklist

### DIFF
--- a/mythtv/libs/libmyth/mythcontext.cpp
+++ b/mythtv/libs/libmyth/mythcontext.cpp
@@ -440,9 +440,8 @@ bool MythContext::Impl::FindDatabase(bool prompt, bool noAutodetect)
 
     // 1. Either load XmlConfiguration::k_default_filename or use sensible "localhost" defaults:
     bool loaded = LoadDatabaseSettings();
-    DatabaseParams dbParams = GetMythDB()->GetDatabaseParams();
-    setLocalHostName(dbParams.m_localHostName);
-    DatabaseParams dbParamsFromFile = dbParams;
+    const DatabaseParams dbParamsFromFile = GetMythDB()->GetDatabaseParams();
+    setLocalHostName(dbParamsFromFile.m_localHostName);
 
     // In addition to the UI chooser, we can also try to autoSelect later,
     // but only if we're not doing manualSelect and there was no
@@ -511,7 +510,10 @@ bool MythContext::Impl::FindDatabase(bool prompt, bool noAutodetect)
                     manualSelect = false;
                     break;
                 case BackendSelection::kCancelConfigure:
-                    goto NoDBfound;
+                {
+                    LOG(VB_GENERAL, LOG_DEBUG, "FindDatabase() - failed");
+                    return false;
+                }
             }
         }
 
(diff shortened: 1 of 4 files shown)


bennettpeter:
With this pull request - 

Running remote mythfrontend
If there is an empty .mythtv directory, after the Search button is requested, there is a prompt for language and country before you get the list of backends. Even if I select Netherlands and Dutch, everything continues in English, so it is unnecessary and achieves nothing. There is an existing backend database whcih already has language selected.
After the unnecessary language prompt it proceeds correctly, giving the list of backends and letting me enter the pin and connect.

Running mythtv-setup with an empty directory prompts for language three times before exiting. After that is is OK.

There should only be one prompt for language in mythtv-setup and none in mythfrontend.

FYI it  is possible to run many configurations on one computer. This could be used to test the remote frontend case with only one computer. See https://github.com/MythTV/packaging/tree/master/deb-light#developing-and-testing . You could also test remote frontend using two login ids, one for backend and one for frontend.

ulmus-scott:
I did not break LanguageSelection::Save() but that should make it actually work so the translation loads the language requested and does not ask more than once per application run.

I disagree that mythfrontend should not ask for the language.  For example, when setting up a remote frontend, it has not yet connected to the database to retrieve the setting or it is on a new host where there is not yet a value in the database.

bennettpeter:
I could agree with the language prompt if it worked.

After I applied  the latest pull request - 

- I set my backend language to Netherlands - Dutch using the web app and restarted the backend.
- I cleared out the .mythtv directory on my remote frontend
- I started the remote frontend. 
- It prompts in ENGLISH - Unable to connect to database and shows 4 buttons.
- I select the Search button and enter.
- It prompts for Country and Language. I select Netherlands - Dutch (same as the backend is set up for).
- I hit Save.
- It promts in ENGLISH Cannot login to database, prompts me in ENGLISH to enter the backend PIN.
- I enter the pin and hit OK.
- I get mythfrontend with menu displayed in ENGLISH, excpet the date at the top of the screen is in Dutch
- I try a few screens, everything else is in ENGLISH.
- Exit frontend (exit menu is in ENGLISH)
- Restart frontend, menu is now in Dutch.
- I try a few screens, everything is in Dutch.

Having people use a language prompt that does nothing is not useful. Do we take out the language prompt or make it work? I recommend taking it out.

Allthe rest is working correctly now.

ulmus-scott:
Does mythtv-setup no longer ask for the language multiple times?

https://github.com/MythTV/mythtv/blob/ba105cdb46bc6d63f4b5c05804e46e2bc03a697a/mythtv/libs/libmythbase/mythtranslation.cpp#L38 MythTranslation::load() does not load a new language if the setting has changed, which MythContext assumes it does.

I think the localization should go like this:

1. Load translation file based on OS locale.
2. Connect to database.
2.a Prompt for database parameters in OS language if necessary.
3. Read language setting from database and load translation file for that language.

Then the language prompt is not needed and the language setting can still be changed via mythfrontend->settings->appearance->localization.

bennettpeter:
mythtv-setup prompts for language, I selected Nederlands, the prompts for language again, Nederlands already selected, then exits.
On starting mythtv-setup again, goes into setup entirely in English, apart from the date at the top of the screen.
Exit mythtv-setup and relaunch. Now it is displaying in Dutch.

mythtv-setup is deptrecated so it is not important for it to work perfectly.

ulmus-scott:
What log messages for "Loading %1 translation for module %2" do you see?

ulmus-scott:
I think my change to `LanguageSelection::Save()` actually makes the language selection specific to a hostname instead of global, which may or may not be desired.

The multiple showings of the language prompt may be due to MythContext calling MythDB::ClearSettingsCache(), but I am not sure.

---

Regardless, the LanguageSelection issue is separate from the issue with config.xml that the first commit addresses.

bennettpeter:
I tested it. It is working now with the language prompt and using the language that I selected.

bennettpeter:
Rebased and merged.

PR #1227, opened 2025-12-20
Fixed channel icon conflict when populating the database from XMLTV

janbar:
Based on the remote filenames of the icons, conflicts can arise, causing a mess:

    <icon src="https://static.te.../stations/308/icon320_light.png?v2023_48_0"/>
    <icon src="https://static.te.../stations/58/icon320_light.png?v2023_48_0"/>
    <icon src="https://static.te.../stations/383/icon320_light.png?v2023_48_0"/>

These cases generates the same icon filename 'icon320_light.png?v2023_48_0'.

The proposed solution no longer uses the original filename; it now generates a user-friendly name from the source identifier of channel (xmltvid), followed by the hash of the full remote URL and, optionally, the file extension. This way, the generated name can no longer conflict with that of another channel, and it is easily identifiable with the xmltv identifier, as it is user-configured.

    <xmltv_id>_<remote_url_hash>.<ext>

Now downloaded icon names from xmltv source look like the following.

    T18.fr_2fcaf706.svg.png
    NOVO19.fr_d890da17.svg.png
    TMC.fr_79bd517.png

##### Checklist

- [ X ] contribution does not duplicate one of our [existing pull requests](https://github.com/MythTV/mythtv/pulls)
- [ X ] contribution is in a branch rebased against [master](https://github.com/MythTV/mythtv)
- [ X ] code compiles successfully without errors
- [ X ] code follows the [MythTV Coding Standards](https://www.mythtv.org/wiki/Coding_Standards)
- [ X ] documentation added/updated/removed where necessary
- [ X ] commits are logically organised and have [good commit messages](https://chris.beams.io/posts/git-commit)

### DIFF
--- a/mythtv/programs/mythfilldatabase/channeldata.cpp
+++ b/mythtv/programs/mythfilldatabase/channeldata.cpp
@@ -211,8 +211,13 @@ void ChannelData::handleChannels(int id, ChannelInfoList *chanlist) const
 
         if (!(*i).m_icon.isEmpty())
         {
-            QDir remotefile = QDir((*i).m_icon);
-            QString filename = remotefile.dirName();
+            QString remotename = QDir((*i).m_icon).dirName();
+            // Define an unique icon filename for this channel and xmltv source
+            // The most obvious format is <source_id>_<xmltv_id>_<remotename>
+            QString filename(QString::number(id));
+            filename.append("_").append((*i).m_xmltvId);
+            filename.append("_").append(remotename);
+            filename.remove(QRegularExpression("[ \\\\/\\:\\*\\[\\]\\(\\)]"));
 
             localfile = fileprefix + filename;
             QFile actualfile(localfile);


garybuhrmaster:
While I understand your particular issues,this feels like an incomplete (and therefore wrong) solution.

In my case, for one guide source, the base names are entirely meaningful and useful (includes the channel and network association), and hiding them inside some opaque hash would be a disservice (perhaps unless you want to go whole hog and create a unique uuid based channel icon blob table which has automatic deletion when there are no longer any references, but that is probably a bridge too far).

And the xmltv is only unique within a sourceid (when you have multiple guide sources), as the same xmltv (which is defined only by the source) can be reused by other sourceids (yes, there is the RFC2838 standard,but not all guide sources use it).

Instead, I would suggest that as the xmltv will be only unique within the sourceid, that the (default) names should be something of the form of sourceid_xmltv_base_url_name.

janbar:
@garybuhrmaster , I amended the initial commit. Thanks for review.

janbar:
Hi, have you any news for this fix ?

Currently I have to compile my own branch including this fix to work around the issue. Is it could be merged or abandoned ?

bennettpeter:
I will test it and if it looks good I will merge it into master

bennettpeter:
With this change, on a run of mythfilldatabase a full set of new icons is downloaded and added to the channels directory. The old ones are not deleted. Previously my channels directory had 195 icons, now it has 398 icons. 203 files were added, so I assume that I already had a few cases of more than one channel using the same icon file name.

File names look like this now, for my case
1_I10021.json.schedulesdirect.org_s10021_dark_360w_270h.png
1_I11000.json.schedulesdirect.org_GNLZZGG003PD8JJ.png_light_360w_270h.png

The no longer used icons are occupying 4 MB of disk space. They could easily be deleted as they are the ones with names that do not start with source id and xmltv id

janbar:
Good to hear !
I'm not the only one whose icons are messed up. The fix is useful.

bennettpeter:
I have merged it into master.

PR #1228, opened 2025-12-20
Minor cleanup to Goom code

ulmus-scott:
Fix some includes, use `<cstdint>`, and use MythRandom instead of rand().

##### Checklist


ulmus-scott:
Should I also convert tabs to spaces in the goom files?

linuxdude42:
I would say not, in case we ever want to try and import a newer version of goom. Although that's not very likely at this point.

ulmus-scott:
Regarding a newer version of goom, is there any?  And at this point, I think it has been rather modified to fit in with MythTV.

The best source I could find (which I put on https://wiki.mythtv.org/wiki/External_libraries) was https://sourceforge.net/projects/goom/ which lists the last update as 2013-04-03.

PR #1253, opened 2026-01-10
Add more GitHub Actions builders

ulmus-scott:
I intend to add an aarch64 Linux build and Android and MinGW cross compiled builds.

### DIFF
--- a/.github/workflows/build-mythtv.yml
+++ b/.github/workflows/build-mythtv.yml
@@ -62,8 +62,6 @@ jobs:
   build-mythtv-linux:
     name: Building MythTV on ${{ matrix.container.desc }}
 
-    runs-on: ubuntu-latest
-
     env:
       TZ: Etc/UTC
       CCACHE_DIR: $HOME/.ccache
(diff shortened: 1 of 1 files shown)


garybuhrmaster:
This may be a bad idea in practice (although a good idea in theory).

The concern is that github has a limit on the total cache size for the project.  The existing limit on the number of runners was chosen to fit within that limit (with a bit of breathing room for expected code growth), so that the actions can be completed using the ccache of the runners and the actions can complete is a reasonable time frame.

So the question would be what was the result of your testing to insure that the total github cache sizes are still useful and left some additional headroom?

ulmus-scott:
The size of the ccache is limited to 250 MB, with some additional compression, the caches end up being around 220 MB in size.

11 * 220 MB = 2.42 GB; 14 * 220 MB = 3.08 GB

So there is still room for three separate runs.

However, if saving the cache for pull requests is not desired, we could use separate [restore](https://github.com/actions/cache/blob/main/restore/README.md) and [save](https://github.com/actions/cache/blob/main/save/README.md) cache steps with the save cache step using `if: github.ref == 'refs/heads/master'` and then we could use all 10 GB at once.

`if: '!canceled()' && github.ref == 'refs/heads/master'` might be a better condition, so the cache is uploaded even if compilation or testing fail.

---

NB -- The ccache has a higher miss rate for cmake Fedora.  Perhaps due to using LTO?

ulmus-scott:
Test remove caches actions.

PR #1255, opened 2026-01-12
 Add more GitHub Actions builders

ulmus-scott:
I intend to add an aarch64 Linux build and Android and MinGW cross compiled builds.

---

See also comments from https://github.com/MythTV/mythtv/pull/1253

I think I should have reopened the pull request before force pushing to it.

### DIFF
--- a/.github/workflows/build-mythtv.yml
+++ b/.github/workflows/build-mythtv.yml
@@ -66,7 +66,7 @@ jobs:
       TZ: Etc/UTC
       CCACHE_DIR: $HOME/.ccache
       CCACHE_COMPRESS: true
-      CCACHE_MAXSIZE: 250M
+      CCACHE_MAXSIZE: 300M
       DEBIAN_FRONTEND: noninteractive
 
     strategy:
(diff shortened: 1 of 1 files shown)


ulmus-scott:
Cache hit rate is higher than I thought:

- Linux configure: about 92.5%, now just over 250M
- Fedora CMake: about 80%, now 300M
- macOS: 100%, just under 250M

stuarta:
It does not look correct to use `ubuntu-latest` runners for non ubuntu builders

garybuhrmaster:
The MythTV buildbot network is the official CI.  The github runners are just advisory as the quick and dirty testing.

I am not a decider, but if you want to actually usefully use the few github runners (left with a ccache that can be supported under the github total cache limits), I would create runners that run clazy and tidy to prevent stupid PRs that only make things worse (i.e. PR's that introduce new issues will be marked as such and rejected immediately).  A number of years ago I looked at such runners and found that the recipe for the running of such in the official buildbot network was not complete/sufficient for github runners (that may have been my misunderstandings, of course).

ulmus-scott:
@stuarta `build-mythtv-linux` is run in Docker containers, so I don't think the host, which is unchanged, is that important.

@garybuhrmaster is all of the buildbot stuff at https://github.com/MythTV/buildbot-utils?

billmeek:
This: ansible_cmd: 'ansible-playbook --limit localhost -e "qt6=true" mythtv.yml'
selects qt6 OK, but not because -e "qt6=true". "qt6=false" (or any string will
evaluate to true.)  The string argument may stop working in the future.

This would pass true as a bool:

  ansible-playbook --limit localhost --extra-vars "{'qt6':true}" mythtv.yml

I can provide additional detail if needed. [just -e is OK]

ulmus-scott:
@billmeek It now uses the correct parameter passing syntax for a Boolean.  I had just copied what was already in the file.

PR #1256, opened 2026-01-14
Remove broken WebFrontend

ulmus-scott:
This has been broken since the removal of the
old service API code from mythbackend on 2024-12-09 in 077327d92078ff3019f165e5cbe4ef01675d861c.

It was replaced by the web application.

---

@bennettpeter I don't think I removed anything used by the new web application, but could you check that?  I'm also not sure if any more files are unused.

##### Checklist


stuarta:
I would check if this breaks the upnp functionality for serving media to remote devices or not

bennettpeter:
I tested UPNP -
- the UPNP detection of the backend by a newly installed frontend:
  It says it cannot open the database, then if you select search it finds a list with 1 backend but then says it cannot find a backend. If you select the 1 backend that it did find, it connects successfully. This happens with master as well as with this pull request. It should not complain about the database and should go straight to the backend found if there is one single backend found and the PIN is 0000. This (minor) bug is unrelated to the pull request, as it behaves the same with or without the pull request
- the UPNP detection of media items, tested with VLC, works perfectly with this pull request.

ulmus-scott:
This also removes the ability to download files in a storage group using a `http://<backend host>:6550/StorageGroup/<storage group name>/<file name>` URL, but I suspect that was intended for showing the preview pictures not actually downloading the files.

The only other thing HtmlServerExtension does is run QtScript files and host files in `share/mythtv/html`.

bennettpeter:
You can now use http://<backend>/Content/GetFile?StorageGroup=xxxx&FileName=yyyy to download files.
Are there any places in the code where the old URL is used?

bennettpeter:
I have fixed the bug mentioned above in UPNP frontend startup. See 67ca623b58

ulmus-scott:
@bennettpeter The only place where I could find a URL like `http://<backend host>:6550/StorageGroup/<storage group name>/<file name>` is in https://github.com/MythTV/mythtv/blob/67ca623b581793dddb2df6fa96d6c1e30181aac9/mythtv/libs/libmythtv/HLS/httplivestream.cpp#L750-L767

but the port used there is 6544 not 6550.

bennettpeter:
The livestream methods were not implmented in the new V2 Service API.
There were methods like this before: 
http://backend:6544/Content/AddRecordingLiveStream?RecordedId=13&MaxSegments=20
http://backend:6544/Content/GetLiveStream?Id=9
I tested the old methods and they did not work. I assume they have not been used for a long time. So I did not implement them in the new API.
Probably the livestream can be removed from MythTV. I believe it was originally used with mythweb but the playback feature of mythweb stopped working some years ago with the demise of certain browser plugins.

ulmus-scott:
`httplivestream.(h|cpp)` are currently only used for the --hls mode in mythtranscode.

ulmus-scott:
The old WebFrontend is the last real use of QtScript.  Should I add the commit to remove the HtmlServerExtension from mythfrontend here, or should that be a separate pull request?

bennettpeter:
Probably should be separate. I will merge this request now.

bennettpeter:
wsdl display no longer works correctly, as it is looking for wsdl.css that is now deleted.
I tried copying wsdl.css back in but the server still does not find it.

ulmus-scott:
I think you would need to add back the `"/css/"` directory to
https://github.com/MythTV/mythtv/blob/acd1265db0b6d1f755a7feb8b5805ea7bd07816b/mythtv/libs/libmythbase/http/mythhttpserver.cpp#L60

or maybe add it back as `mythtv/html/xslt/css/wsdl.css`?

```
git grep -nE "(Status|settings|setup|site|wsdl).css"
mythtv/html/xslt/class.xslt:9:          <link href="/css/wsdl.css" rel="stylesheet" type="text/css" />
mythtv/html/xslt/enum.xslt:11:          <link href="/css/wsdl.css" rel="stylesheet" type="text/css" />
mythtv/html/xslt/service.xslt:10:<link href="/css/wsdl.css" rel="stylesheet" type="text/css" />
mythtv/programs/mythbackend/httpconfig.cpp:382:       << "  <link rel=\"stylesheet\" href=\"/setup/css/Config.css\" type=\"text/css\">\r\n"
mythtv/programs/mythbackend/httpstatus.cpp:644:       << "  <link rel=\"stylesheet\" href=\"/css/Status.css\" type=\"text/css\">\r\n"
mythtv/programs/mythbackend/servicesv2/v2status.cpp:840:       << "  <link rel=\"stylesheet\" href=\"/css/Status.css\" type=\"text/css\">\r\n"
```

It looks like `css/Status.css` is also used.  I'm sorry I missed that those two files were used.

There was no `setup/css/Config.css` for me to remove , so I don't know what `mythbackend/httpconfig.cpp` is trying to reference.

bennettpeter:
I will add back the css directory and the entries that reference it. We need to have it in place for the V36 release next week.

ulmus-scott:
`setup/css/Config.css` was removed in e9ed458c8dd45481c856e28bc9b1a9d93d3801fb in 2011.

@linuxdude42 The only file in https://ftp2.osuosl.org/pub/mythtv/3rdParty/ is https://ftp2.osuosl.org/pub/mythtv/3rdParty/jwplayer.zip which is from 2012-01-16.

This would only be accessed by

https://github.com/MythTV/mythtv/blob/c84e3c805fde78da0c495dc9dfab31ea041245f2/mythtv/programs/mythbackend/httpconfig.cpp#L158-L159

, but `jwplayer` occurs nowhere in the repository anymore, so I think that file can be removed.

@bennettpeter Since the web application exists now, can `mythbackend/httpconfig.*` be removed?  I couldn't see if it actually works since it was asking for a login http://localhost:6550/Config etc., but I don't know what it would be.

ulmus-scott:
Editing [HTTPRequest::IsUrlProtected()](https://github.com/MythTV/mythtv/blob/c84e3c805fde78da0c495dc9dfab31ea041245f2/mythtv/libs/libmythupnp/httprequest.cpp#L1802) to always return false, I can see http://localhost:6550/Config/General and http://localhost:6550/Config/Database.

The only setting I did not see in the web application was `Your local time zone (for XMLTV)` under Locale Settings in Config/General.

I think the CSS has been missing/broken since https://github.com/MythTV/mythtv/commit/e9ed458c8dd45481c856e28bc9b1a9d93d3801fb in 2011.

bennettpeter:
I don't know what the passsword is. I think that config is part of the old web frontend that was never completed, so can be removed.

I committed the changes to restore the css but I did not do the pro file. Is that one necessary?

ulmus-scott:
Yes, the change to html.pro is necessary.  Without it, the css folder is not installed when using configure and qmake, so http://localhost:6544/Content/wsdl does not have the CSS styling.

By the way, the username admin and password mythtv worked when I tried it.

PR #1281, opened 2026-01-30
CMake: Build and install libmythdvdnav before libmythtv

ulmus-scott:
The wrong version.h (i.e. libmythbase/version.h) was getting included, causing unnecessary recompilations of avformatdecoder.cpp (transitive include from one of the libmythtv/DVD/ headers).

I first moved the forward declarations of the MythTV IO wrappers used by libmythdvdnav into libdvdread to remove the header dependency on libmythtv.

Installing libmythdvdnav first also revealed use of internal functions from libdvdread.

##### Checklist


ulmus-scott:
@linuxdude42 I think I did the CMake subproject correctly, but it looks like the libudfread project doesn't set the compiler flags to the same as the rest of MythTV.

Issue #1290, opened 2026-02-06
UPnP discovery of SatIP network tuner fails


kmdewaal:
* **Platform**: Ubuntu 24
`Linux Carbon8 6.8.0-94-generic #96-Ubuntu SMP PREEMPT_DYNAMIC Fri Jan  9 20:36:55 UTC 2026 x86_64 x86_64 x86_64 GNU/Linux
`
* **MythTV version**: master and v36

* **Package version**: compiled from source

* **Component**: mythbackend / UPnP / SatIP

### What steps will reproduce the bug?

Start mythbackend.
This will then do a test on all configured tuners and this includes a UPnP discovery of all SatIP network tuners. This works correct on v35 but it fails on v36 and on master.
This is a part of the v36 mythbackend log that shows the discovery failure:
```
2026-02-06 12:02:42.533198 I [392831/392831] CoreContext satiputils.cpp:30:probeDevices  SatIP: Using UPNP to search for Sat>IP servers (3 secs)
ST: urn:ses-com:device:SatIPServer:1
2026-02-06 12:02:43.036346 D [392831/392831] CoreContext satiputils.cpp:45:probeDevices  SatIP: UPNP search 2 ms
ST: urn:ses-com:device:SatIPServer:1
2026-02-06 12:02:43.519957 D [392831/392831] CoreContext satiputils.cpp:45:probeDevices  SatIP: UPNP search 2 ms
ST: urn:ses-com:device:SatIPServer:1
2026-02-06 12:02:43.864826 D [392831/392831] CoreContext satiputils.cpp:45:probeDevices  SatIP: UPNP search 1 ms
ST: urn:ses-com:device:SatIPServer:1
2026-02-06 12:02:44.345910 D [392831/392831] CoreContext satiputils.cpp:45:probeDevices  SatIP: UPNP search 1 ms
ST: urn:ses-com:device:SatIPServer:1
ST     =urn:ses-com:device:SatIPServer:1
USN    =uuid:59e26221-0631-4539-bb32-2253db5d1b1c::urn:ses-com:device:SatIPServer:1
2026-02-06 12:02:45.785224 E [392831/392831] CoreContext satiputils.cpp:77:doUPNPsearch  SatIP: No UPnP Sat>IP servers found, but SSDPCache::Instance()->Find() != NULL
```

### How often does it reproduce? Is there a required condition?
100% reproducible.

### What is the expected behaviour?
Detection of the SatIP network tuners.
[...]

PR #1293, opened 2026-02-08
Always process events in PortChecker

ulmus-scott:
Other threads have QEventLoops, so always call
QCoreApplication::processEvents().  Do not add
unnecessary sleep_for() calls, but yield the thread's time slot, so other threads have a chance to run.

This reverts commit f50bc3a088a1b93dfee4a2d5614105c071729922.

Revert "PortChecker: Only processEvents() in the UI thread"

This reverts commit b85319e0e0c29f08856b65ffe61858c48c099f62.

---

@kmdewaal This should resolve https://github.com/MythTV/mythtv/issues/1290.  I don't know if we still want to only try connecting for a maximum of three seconds before restarting the connection attempt, so I have restored that behavior.

##### Checklist

### DIFF
--- a/mythtv/libs/libmythbase/portchecker.cpp
+++ b/mythtv/libs/libmythbase/portchecker.cpp
@@ -29,7 +29,6 @@
 #include <QCoreApplication>
 #include <QHostAddress>
 #include <QTcpSocket>
-#include <QThread>
 #include <QEventLoop>
 #include <QNetworkInterface>
 #include <QNetworkAddressEntry>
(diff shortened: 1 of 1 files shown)


kmdewaal:
About restarting the connection after 3 seconds. I think that does not make sense at all; if you want to wait 5 seconds for the connection to happen, which is theoretically possible for a very slow device and/or network, it does not make sense to abort the attempt after 3 seconds and then do another one for only two seconds.
But, as on my system the port check is done in 20ms I do not worry about it.

kmdewaal:
@linuxdude42 @ulmus-scott I have merged and tested this PR and it works OK. Have also committed it to v36 and tested that as well so I think we should be good.
Thanks for the support and I am happy that SatIP is still working now.

PR #1312, opened 2026-02-26
Have 'make uninstall' delete plugins which are no longer built

### LINKED ISSUE #1174: Plugin with mismatched version causes crash
* **Platform**: Linux Mint 22.1, 5.15.0-124-generic 134-Ubuntu SMP Fri Sep 27 20:20:17 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux, Qt version 5.15.13

* **MythTV version**: 36

* **Package version**:

* **Component**: mythfrontend, mythweather

Install an old plugin, such as **mythweather** (v34), on a system running a newer release (v36). 
```
$ ls -l /usr/lib/mythtv/plugins
total 36588
-rwxr-xr-x 1 root root  3933320 Aug 15 20:41 libmytharchive.so*
-rwxr-xr-x 1 root root  1513464 Aug 15 20:41 libmythbrowser.so*
-rwxr-xr-x 1 root root 16515296 Aug 15 20:41 libmythmusic.so*
-rwxr-xr-x 1 root root  4359712 Aug 15 20:41 libmythnews.so*
-rwxr-xr-x 1 root root 11129656 Aug 22 17:31 libmythweather.so*
$ strings /usr/lib/mythtv/plugins/* | grep libmythbase-
libmythbase-36.so.36
libmythbase-36.so.36
libmythbase-36.so.36
libmythbase-36.so.36
libmythbase-34.so.34
$
``` 
Run **mythfrontend**. In the standard output, you should see something like this:
```
2025-08-11 23:55:03.961399 ! [609542/609542] CoreContext mythcorecontext.cpp:2117:TestPluginVersion  MythCoreContext::TestPluginVersion(): Plugin mythweather (34.20220913-1) binary version does not match libraries (36.20250212-1)
2025-08-11 23:55:03.961406 E [609542/609542] CoreContext mythplugin.cpp:143:init_plugin  Unable to initialize plugin 'mythweather'.
```
Next, try to exit from **mythfrontend**. Upon shutdown operations, there will be a crash. For example:
```
2025-08-11 23:55:29.801188 I [609542/609542] CoreContext mythcontext.cpp:1698:~MythContext  Waiting for threads to exit.
2025-08-11 23:55:30.169564 I [609542/609542] CoreContext mythcontext.cpp:1704:~MythContext  Exiting
double free or corruption (!prev)
Aborted (core dumped)
$ coredumpctl gdb -1
...
(gdb) bt
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=<optimized out>, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x000077edcec4527e in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x000077edcec288ff in __GI_abort () at ./stdlib/abort.c:79
#5  0x000077edcec297b6 in __libc_message_impl (fmt=fmt@entry=0x77edcedce8d7 "%s\n") at ../sysdeps/posix/libc_fatal.c:134
#6  0x000077edceca8ff5 in malloc_printerr (str=str@entry=0x77edcedd1ae0 "double free or corruption (!prev)")
    at ./malloc/malloc.c:5772
#7  0x000077edcecab12c in _int_free_merge_chunk (av=0x77edcee03ac0 <main_arena>, p=0x5d2a2fa799c0, size=160)
    at ./malloc/malloc.c:4679
#8  0x000077edcecab43a in _int_free (av=0x77edcee03ac0 <main_arena>, p=<optimized out>, have_lock=<optimized out>)
    at ./malloc/malloc.c:4646
#9  0x000077edcecaddae in __GI___libc_free (mem=0x5d2a2fa799d0) at ./malloc/malloc.c:3398
#10 0x000077edd1863276 in QTypedArrayData<unsigned short>::deallocate (data=0x5d2a2fa799d0)
    at /usr/include/x86_64-linux-gnu/qt5/QtCore/qarraydata.h:239
#11 0x000077edd1862947 in QString::~QString (this=0x77edd1bab1f8 <Parse::kMythRSS>, __in_chrg=<optimized out>)
    at /usr/include/x86_64-linux-gnu/qt5/QtCore/qstring.h:1308
#12 0x000077edcec47382 in __cxa_finalize (d=0x77edd1ba5e00) at ./stdlib/cxa_finalize.c:82
#13 0x000077edd1860d57 in __do_global_dtors_aux () at /lib/libmythbase-36.so.36
#14 0x00007ffc743498d0 in ??? ()
#15 0x000077edd2be00f2 in _dl_call_fini (closure_map=0x77edd2bddbc0) at ./elf/dl-call_fini.c:43
(gdb)
```
The crash back trace points to code automatically generated by the compiler to run destructors for global or static memory variables. 

It happens every time. The required condition is that there was an attempt to load a plugin built for a non-matching MythTv version.

An old plugin should not cause a crash, but should just be ignored, after giving an error message.

It seems that loading the old plugin causes damage to memory cleanup operations. Can you safely unload it to remove the damage? Might it be possible to check the versioning before trying to load it?

Alternatively, perhaps the Makefile for plugins could be updated to diagnose this condition? When the builder runs `sudo make install`, upon completion of copying plugins to the target directory, it could look for any existing plugins in /usr/lib/mythtv/plugins which have an non-matching version? If make kicked out a fatal error, it would help prevent arriving at this dangerous mix of versions in the first place.
### What do you see instead?

Attempting to load an old plugin seems to mess up C++ destructor handling. Maybe this is because the old plugin was linked against an old version of a shared library (e.g. libmythbase)? The sizes of some structures have changed between versions of this library, so the memory cleanup makes mistakes.

This leads to odd memory handling crashes which are hard to track to specific MythTv code. In #1159 it was found that trying to modify the Theme would lead to a crash. This is because damaged destructors for the graphics context with the old theme are invoked before creating a new graphics context using the new theme.

Sometimes these crashes are blamed on _**double free or corruption (!prev)**_ and sometimes it just looks like a Segmentation Error.

SteveErl:
This update modifies the **uninstall** rule in mythplugins/cleanup/Makefile to delete plugins which are not being built. If a builder reruns configuration to eliminate a plugin which was previously included, but fails to run uninstall before doing so, the old plugin may be left behind. Running `make uninstall` after the configuration changes wouldn't work. These changes will remove plugins which are no longer configured to be built.

Resolves #1174

##### Checklist

### DIFF
--- a/mythplugins/configure
+++ b/mythplugins/configure
@@ -953,21 +953,18 @@ check:
 all:
 clean:
 distclean:
+install:
 qmake_all:
 	\$(NONE)
 
-# Remove any old plugins which we are no longer building. This helps
-# prevent memory corruption caused by mismatches between library
-# versions at the time the old plugin was built and the current
-# versions of those libraries.
-install:
-	rm -f \$(foreach slib, $unplugs, \\
-            \$(INSTALL_ROOT)/$unslashprefix/$libdir_name/mythtv/plugins/$SLIBPREF\$(slib)$SLIBSUF)
-
 # Hack to remove some empty directories that qmake doesn't create rules for
+# Also, remove any old plugins which we are no longer building, but
+# might have been left behind.
 uninstall:
 	-rmdir \$(INSTALL_ROOT)/$unslashprefix/share/mythtv/*
 	-rmdir \$(INSTALL_ROOT)/$unslashprefix/share/mythtv
+	rm -f \$(foreach slib, $unplugs, \\
+            \$(INSTALL_ROOT)/$unslashprefix/$libdir_name/mythtv/plugins/$SLIBPREF\$(slib)$SLIBSUF)
 
 END_CLEANUP
 
--- a/mythplugins/configure
+++ b/mythplugins/configure
@@ -344,6 +344,7 @@ cc="gcc"
 python="python3"
 
 targetos=`uname -s`
+unplugs=""
 
 if test x"$1" = x"-h" -o x"$1" = x"--help" ; then
 cat << EOF
@@ -613,6 +614,11 @@ WEBENGINE=$(cat mythconfig.mak | grep -e "^CONFIG_QTWEBENGINE=yes")
 EXTRALIBS=$(cat mythconfig.mak | grep -e "^EXTRALIBS=")
 EXTRALIBS=${EXTRALIBS#EXTRALIBS=}
 
+SLIBPREF=$(cat mythconfig.mak | grep -e "^SLIBPREF=")
+SLIBPREF=${SLIBPREF#SLIBPREF=}
+SLIBSUF=$(cat mythconfig.mak | grep -e "^SLIBSUF=")
+SLIBSUF=${SLIBSUF#SLIBSUF=}
+
 MYTHTV_VERSION_MAJMIN=$(cat mythconfig.mak | grep -e "^MYTHTV_VERSION_MAJMIN=")
 MYTHTV_VERSION_MAJMIN=${MYTHTV_VERSION_MAJMIN#MYTHTV_VERSION_MAJMIN=}
 MYTHTV_LIBVERSION=$(cat mythconfig.mak | grep -e "^MYTHTV_LIBVERSION=")
@@ -812,6 +818,7 @@ if enabled archive ; then
   echo "SUBDIRS += mytharchive" >> ./config.pro
 else
   echo "        MythArchive    plugin will not be built"
+  unplugs="archive"
 fi
 
 if enabled browser ; then
@@ -819,6 +826,7 @@ if enabled browser ; then
   echo "SUBDIRS += mythbrowser" >> ./config.pro
 else
   echo "        MythBrowser    plugin will not be built"
+  unplugs="$unplugs browser"
 fi
 
 if enabled game ; then
@@ -826,6 +834,7 @@ if enabled game ; then
   echo "SUBDIRS += mythgame" >> ./config.pro
 else
   echo "        MythGame       plugin will not be built"
+  unplugs="$unplugs game"
 fi
 
 if enabled music ; then
@@ -833,6 +842,7 @@ if enabled music ; then
   echo "SUBDIRS += mythmusic" >> ./config.pro
 else
   echo "        MythMusic      plugin will not be built"
+  unplugs="$unplugs music"
 fi
 
 if enabled netvision ; then
@@ -840,6 +850,7 @@ if enabled netvision ; then
   echo "SUBDIRS += mythnetvision" >> ./config.pro
 else
   echo "        MythNetvision  plugin will not be built"
+  unplugs="$unplugs netvision"
 fi
 
 if enabled news ; then
@@ -847,6 +858,7 @@ if enabled news ; then
   echo "SUBDIRS += mythnews" >> ./config.pro
 else
   echo "        MythNews       plugin will not be built"
+  unplugs="$unplugs news"
 fi
 
 if enabled weather ; then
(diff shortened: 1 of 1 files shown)


linuxdude42:
This does not belong in a "make install" command.  I have no objection to it being part of a "make uninstall" command.

kmdewaal:
As I understand it this patch does remove the old plugins when there is a new plugin built.
But you can have an old build with more plugins than the new build and then there will be one or more old plugins still remaining, which can then cause a crash.
This is another reason to put this functionality in a "make uninstall" and I assume it can then uninstall everything that was previously installed.

SteveErl:
I could try moving it to `make uninstall`, but then it will only work if builders know that there is this sneaky hidden problem and that they need to run `make uninstall` to fix it. Do you always run `make uninstall` before running `make install` in the plugins directory? I haven't been doing that. I just tried it out to see how well it works. The output is a bit scary.

It tries, but luckily fails to remove _/usr/bin/_,  _/usr/share/mythtv/_, _/usr/share/mythtv/bindings_, _/usr/share/mythtv/fonts_, _/usr/share/mythtv/locales_, _/usr/share/mythtv/metadata_, _/usr/share/mythtv/mythconverg_backup.pl_, _/usr/share/mythtv/mythconverg_restore.pl_, _/usr/share/mythtv/sql_, _/usr/share/mythtv/themes_. None of these are parts of the plugins. They are components of the mythbackend, mythfrontend, or the operating system.

I prefer to consider `sudo make install` a request to make sure the plugins currently being built are the only ones to be loaded by **mythfrontend**.

kmdewaal:
A few comments.
- I think this is more a package management problem. It should be possible to build a package and then install that package. I think that some users do actually do this. This then solves the problem of removing the old package.
- For my development I do use "make install" but I use dedicated target directories. Afte rmajor changes I just remove the complete target directory (e.g. /home/klaas/mythtv-master/usr/...) before doing the make install.
- A proper solution would be that mythfrontend does do a better job of checking the version of the plugin before loading it. There some form of checking, as loading plugins sometimes fails with a version number mismatch message, but it also sometimes crashes. Maybe this check can be improved.

SteveErl:
The way the version check is currently implemented is that the plugin detects that its version does not match the libmythbase version. But this check is only done once the plugin is loaded, and by then, it's too late to avoid the memory corruption.  

I looked into ways to implement your suggested "proper solution". The only way I found was
```
$ strings /usr/lib/mythtv/plugins/libmythmusic.so | grep libmythbase- | sed 's/^.*\.so\.//'
37
$ strings /usr/lib/mythtv/plugins/libmythweather.so | grep libmythbase- | sed 's/^.*\.so\.//'
34
```
This could work on Linux and MacOS, but Windows does not have a direct built-in equivalent that extracts strings from binary files in the same way the Linux _strings_ command does. It also doesn't have a direct built-in equivalent to _sed_.

Implementing your "proper solution" would be quite complicated. The solution I came up with in this Pull Request is tremendously more simple. It will throw out libmythweather.so on my system since that plugin is expecting an old libmythbase version (34). As a result, it will never get loaded, and we avoid the memory corruption. The builder doesn't need to know about a new requirement that they run `sudo make uninstall` before `sudo make install`.

I grant that it's unusual to delete something during a `make install` command, but it's the simplest, most elegant solution to this problem.

SteveErl:
I've moved the cleanup of old plugins from the **install** rule to the **uninstall** rule.

PR #1363, opened 2026-05-07
Prevent segmentation fault when exiting GUIStartup

ulmus-scott:
Without this, the outer while loop will continue infinitely and cause a segmentation fault at the next checkPort() call in the call to QObject::connect, suggesting the GUIStartup instance pointed to by MythContext::Impl::m_guiStartup has been deleted, probably via the call to MythScreenType::Close() in GUIStartup::OnClosePromptReturn().  This might still be a race condition dependent use after free, however.

---

To trigger this start mythfrontend without a backend to connect to and immediately exit the GUIStartup screen once it is shown.

##### Checklist

- [x] commits are logically organized and have [good commit messages](https://chris.beams.io/posts/git-commit)

### DIFF
--- a/mythtv/libs/libmyth/mythcontext.cpp
+++ b/mythtv/libs/libmyth/mythcontext.cpp
@@ -982,8 +982,12 @@ QString MythContext::Impl::TestDBconnection(bool prompt)
             }
             if (m_guiStartup)
             {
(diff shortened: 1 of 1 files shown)


kmdewaal:
Reproduced the problem and tested the fix.

PR #1398, opened 2026-06-24
Add support for ECDSA (and other) TLS certificates

a-a-ron-t:
Wherever a QSslKey is created, use the key algorithm from the certificate instead of hard-coding one. This enables using any type of certificate that Qt supports.

I've tested the MythHTTPS and uPNP HttpServer changes with a local build of fixes/36 on Debian 13, using ECDSA certs from Let's Encrypt. I'm not sure how to test the MHEG change, but it compiles with no issues.

I put the code moves and actual code changes into separate commits to hopefully make things clear, since the changes are so small compared to the moves, but I'm happy to do some squashing if that would be preferable.

##### Checklist

### DIFF
--- a/mythtv/libs/libmythbase/http/mythhttps.cpp
+++ b/mythtv/libs/libmythbase/http/mythhttps.cpp
@@ -88,7 +88,7 @@ bool MythHTTPS::InitSSLServer(QSslConfiguration& Config)
     }
 
     auto rawHostKey = hostKeyFile.readAll();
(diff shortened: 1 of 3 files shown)


linuxdude42:
I have tested the changes to mythhttps.cpp with RSA and ECDSA certificates, and successfully loaded an ECDSA client certificate in netstream.cpp.  (The latter required a small tweak to force the https code path  to execute since my two MHEG test videos only reference http web pages.) I have not had a chance to test the upnp changes, but they appear the same as the mythhttps.cpp changes.

PR #1418, opened 2026-07-24
Update libmythdvdnav

ulmus-scott:
to libdvdread version 7.1.1 at 0b21c45f648a38947ee2bbc75c9a1bd577d96b71 and to libdvdnav version 7.0.1 at e0c02b973c62081ee8dc109726e511e94c10f70e.

NB: dvdnav_read_name() now only works for local DVDs, so if previously the name and serial number of a remote DVD was returned, it no longer will be.

The following commit was reverted:
Fix null pointer dereference in libmythdvdnav.
https://github.com/MythTV/mythtv/commit/de9344c2a8f18439e26e4ed11c85dde95867318a

See https://code.videolan.org/videolan/libdvdread/-/merge_requests/97#note_563739 for why.

libdvdread is minimally modified but the non-exported function InternalUDFReadBlocksRaw() is used in libmythtv/DVD/mythdvdstream.cpp.

libdvdnav still has more extensive modifications.

---

There is no longer a circular dependency between libmythtv and libmythdvdnav.  However, I am uncertain if the remote playback capabilities are the same.

I have tested only local DVD playback.  I do not have a set up for testing the remote DVD playback.  So if anyone could verify what worked before and after this change for remote playback, that would be great.

Also, can someone test the bookmarking/DVD VM state save and restore?

If someone could test whether https://github.com/ulmus-scott/libdvdread/commit/e4e99e7e0d803537975f571b8aaa13b14d8e6bac (from https://github.com/MythTV/mythtv/commit/1ff446ad093064ff365d60561224231b6c3e7c5b) is still necessary, that is the last major change to libdvdread and I can attempt to upstream it if it is necessary.

---

##### Checklist

- [x] commits are logically organized and have [good commit messages](https://chris.beams.io/posts/git-commit)


mythtv/external/libmythdvdnav/CMakeLists.txt
```diff
@@ -149,6 +163,16 @@ if(_NEEDS_DLFCN)
 endif()
 
 check_include_file(sys/param.h HAVE_SYS_PARAM_H)
+check_include_file(dirent.h HAVE_DIRENT_H)
+check_symbol_exists(getmntent_r "mntent.h" HAVE_GETMNTENT_R)
+check_symbol_exists(strerror_r "string.h" HAVE_STRERROR_R)
+check_symbol_exists(strerror_s "string.h" HAVE_STRERROR_S)
+
+# TODO: This doesn't work, but meson does.
```
ulmus-scott: @linuxdude42 I'm not sure why _Static_assert doesn't work in CMake's compile test.  It works in meson and configure.

ulmus-scott:
There is no rush to merge this and I am still working with upstream on some of the changes.

mythtv/external/libmythdvdnav/CMakeLists.txt
linuxdude42: The error is: (.text+0x1b): undefined reference to `main'.

This works:

"int main (int a, char **b) { _Static_assert(1, \"The impossible happened.\"); }"

ulmus-scott: Thanks, that works.  How did you see what the error was?

linuxdude42: its kinda hacky, but here's what I did:
```cmake -S mythtv/external/libmythdvdnav -B build-dvd --debug-trycompile```
Look at the output for the line "Executing try_compile (HAVE_STATIC_ASSERT) in:"
```cd <tempdir>```
```make```

ulmus-scott: I should read the documentation more carefully: https://cmake.org/cmake/help/latest/module/CheckCSourceCompiles.html .  It compiles **and links into an executable**, so would need `main()`.

PR #1449, opened 2026-08-09
CMake: Add --install option to super-project

ulmus-scott:
Add an --install option to the CMake super-project.

If CMAKE_INSTALL_PREFIX is not set, the project will now install the sub-projects inside the created build-* directory under the install directory.

Additionally, if MYTH_RUN_PREFIX is also not set, it will now default to CMake's default CMAKE_INSTALL_PREFIX, likely /usr/local.

If CMAKE_INSTALL_PREFIX was set and MYTH_RUN_PREFIX was not, the install command does nothing.

If MYTH_RUN_PREFIX was set, the install command will copy the install to that directory.

##### Checklist

- [x] contribution does not duplicate one of our [existing pull requests](https://github.com/MythTV/mythtv/pulls)
- [x] contribution is in a branch rebased against [master](https://github.com/MythTV/mythtv)
- [x] code compiles successfully without errors
- [x] code follows the [MythTV Coding Standards](https://www.mythtv.org/wiki/Coding_Standards)
- [x] documentation added/updated/removed where necessary
- [x] commits are logically organised and have [good commit messages](https://chris.beams.io/posts/git-commit)

PR #1451, opened 2026-08-10
CMake: Alternative pull to add support for a "cmake --install" step.

linuxdude42:
Add support for a staging directory to hold the results of the "build" stage, and add support for "cmake --install" command to copy the staging directory to the requested install directory.  The staging directory is enabled by default for unixish builds, and is ignored for android/windows builds.


CMakeLists.txt
```diff
@@ -534,3 +553,11 @@ endif()
 include(platform/${CMAKE_SYSTEM_NAME}${_CROSS}Packaging OPTIONAL)
 
 include(StaticAnalysis)
+
+# Install mythtv.  This does nothing more than copy everything from
+# the temporary install directory to CMAKE_INSTALL_PREFIX.
+if(MYTH_STAGING_PREFIX)
+  install(DIRECTORY ${MYTH_STAGING_PREFIX}/ DESTINATION . USE_SOURCE_PERMISSIONS)
```
ulmus-scott: You need to exclude the pkg-config files, since they are only relevant during the build.  This still creates an empty pkgconfig folder, though:
```
  install(DIRECTORY ${MYTH_STAGING_PREFIX}/ DESTINATION . USE_SOURCE_PERMISSIONS
          PATTERN "ffnvcodec.pc" EXCLUDE # these headers are not installed
          PATTERN "*.pc" EXCLUDE # these point inside MYTH_STAGING_PREFIX
          )
```

ulmus-scott:
I was trying to keep the behavior unchanged in https://github.com/MythTV/mythtv/pull/1449, but I think the semantics of CMAKE_INSTALL_PREFIX are better in yours.

Based on how this works with CMAKE_INSTALL_PREFIX, does this obviate a need for a separate MYTH_RUN_PREFIX option?

CMakeLists.txt
linuxdude42: Given the second pattern, the first pattern it redundant, isn't it?

linuxdude42:
Probably? It appears that its only remaining use is (i think) in building the python package. I'm not a python person. I can leave it or remove it, doesn't matter to me. Maybe leave for now and remove in a future commit after checking with someone fluent in python.

CMakeLists.txt
ulmus-scott: Well, yes, the first pattern is redundant, but the reasoning is different.  The comment could probably be: ffnvcodec is not installed and the other pkg-config files point inside MYTH_STAGING_PREFIX.

ulmus-scott:
I think we should leave MYTH_RUN_PREFIX as is for now.  I don't think it would need to be an option anymore, but I'm not sure the super-project's CMAKE_INSTALL_PREFIX is passed through to the sub-projects otherwise.  The python bindings only need it to find the grabber scripts and the mythtv executables (data_mythtv.py, to run `--version`).

```
git grep -nw INSTALL_PREFIX
mythplugins/mytharchive/README:154:    INSTALL_PREFIX/share/mythtv/ or /etc/mythtv/
mythplugins/mytharchive/README:158:    INSTALL_PREFIX/share/mythtv/mysql.txt
mythtv/bindings/python/MythTV/_versions.py.in:16:INSTALL_PREFIX = '@MYTHTV_INSTALL_PREFIX@'
mythtv/bindings/python/MythTV/dataheap.py:1314:                        prefix=os.path.join(INSTALL_PREFIX, 'share/mythtv'))
mythtv/bindings/python/MythTV/dataheap.py:1342:            long_path = os.path.join(INSTALL_PREFIX, path)
mythtv/programs/scripts/hardwareprofile/distros/mythtv_data/data_mythtv.py:54:            execpath = os.path.join(MythTV.static.INSTALL_PREFIX,
```

